PR #3, opened 2018-01-23
define action protocol

### LINKED ISSUE #2: Create proto file for action plugin
Right now, we only have a few examples for a subset of our plugins.

We need to write the full proto files and review them before we can start implementing the frontends (e.g. DroneCore-Python and DroneCore-Java).

JonasVautherin:
Fixes #2.

### DIFF
--- a/action/action.proto
+++ b/action/action.proto
@@ -7,24 +7,37 @@ option java_outer_classname = "ActionProto";
 
 service ActionService {
     rpc Arm(ArmRequest) returns(ActionResult) {}
-    rpc TakeOff(TakeOffRequest) returns(ActionResult) {}
+    rpc Disarm(DisarmRequest) returns(ActionResult) {}
+    rpc Takeoff(TakeoffRequest) returns(ActionResult) {}
     rpc Land(LandRequest) returns(ActionResult) {}
+    rpc Kill(KillRequest) returns(ActionResult) {}
+    rpc ReturnToLaunch(ReturnToLaunchRequest) returns(ActionResult) {}
+    rpc TransitionToFixedWings(TransitionToFixedWingsRequest) returns(ActionResult) {}
+    rpc TransitionToMulticopter(TransitionToMulticopterRequest) returns(ActionResult) {}
+    rpc GetTakeoffAltitude(GetTakeoffAltitudeRequest) returns(TakeoffAltitudeResult) {}
+    rpc SetTakeoffAltitude(SetTakeoffAltitudeRequest) returns(ActionResult) {}
+    rpc GetMaximumSpeed(GetMaximumSpeedRequest) returns(MaximumSpeedResult) {}
+    rpc SetMaximumSpeed(SetMaximumSpeedRequest) returns(ActionResult) {}
 }
 
-message UUID {
-    uint64 value = 1;
+message ArmRequest {}
+message DisarmRequest {}
+message TakeoffRequest {}
+message LandRequest {}
+message KillRequest {}
+message ReturnToLaunchRequest {}
+message TransitionToFixedWingsRequest {}
+message TransitionToMulticopterRequest {}
+message GetTakeoffAltitudeRequest {}
+
+message SetTakeoffAltitudeRequest {
+    float altitude_m = 1;
 }
 
-message ArmRequest {
-    UUID uuid = 1;
-}
+message GetMaximumSpeedRequest {}
 
-message TakeOffRequest {
-    UUID uuid = 1;
-}
-
-message LandRequest {
-    UUID uuid = 1;
+message SetMaximumSpeedRequest {
+    float speed_m_s = 1;
 }
 
 message ActionResult {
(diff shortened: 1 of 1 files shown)


action/action.proto
```diff
@@ -7,24 +7,37 @@ option java_outer_classname = "ActionProto";
...
 
-message TakeOffRequest {
-    UUID uuid = 1;
-}
+message GetMaximumSpeedRequest {}
 
-message LandRequest {
-    UUID uuid = 1;
+message SetMaximumSpeedRequest {
+    float speed = 1;
```
julianoes: Can we add units? I suggest `speed_m_s` or `speed_mps`.

action/action.proto
```diff
@@ -7,24 +7,37 @@ option java_outer_classname = "ActionProto";
...
+message KillRequest {}
+message ReturnToLaunchRequest {}
+message TransitionToFixedWingsRequest {}
+message TransitionToMulticopterRequest {}
+message GetTakeoffAltitudeRequest {}
 
-message ArmRequest {
-    UUID uuid = 1;
+message SetTakeoffAltitudeRequest {
+    float altitude = 1;
```
julianoes: Units please: `altitude_m`, or `altitudeM`?

JonasVautherin: > "Use CamelCase (with an initial capital) for message names – for example, SongServerRequest. Use underscore_separated_names for field names – for example, song_name. "

From the [guidelines](https://developers.google.com/protocol-buffers/docs/style).

`altitude_m` it is!

action/action.proto
```diff
@@ -37,9 +50,19 @@ message ActionResult {
...
     Result result = 1;
     string result_str = 2;
 }
+
+message TakeoffAltitudeResult {
+    float altitude_m = 1;
+}
+
+message MaximumSpeedResult {
+    float speed_mps = 1;
```
shakthi-prashanth-m: Its should be `speed_m_s` for [consistency ](https://github.com/dronecore/DroneCore/blob/38e89fc6f5f2038b40a7ae4c8414b48914ad1aca/plugins/action/action.h#L255).

JonasVautherin: Fair point. I'll change it tonight!

action/action.proto
julianoes: Ok

action/action.proto
```diff
@@ -7,24 +7,37 @@ option java_outer_classname = "ActionProto";
...
-}
+message ArmRequest {}
+message DisarmRequest {}
+message TakeoffRequest {}
+message LandRequest {}
+message KillRequest {}
+message ReturnToLaunchRequest {}
+message TransitionToFixedWingsRequest {}
+message TransitionToMulticopterRequest {}
+message GetTakeoffAltitudeRequest {}
```
shakthi-prashanth-m: Just For Me, is it necessary to have individual message for each Action request though they're empty ? I mean can't we have [Empty message](https://github.com/google/protobuf/blob/master/src/google/protobuf/empty.proto) ?

JonasVautherin: I am new to protobuf, but I saw this somewhere as a "good practice". The idea is that if you define your protocol saying that it takes an empty message, you will **never** be able to add anything to the request.

If you define a specific message for the request, you could later add parameters to it without breaking the API. And I don't think it hurts performances, as it just generates a few more "empty" messages.

One quick example would be, say, if we wanted to add the "UUID" in all the messages, later. Or something specific to "TransitionToFixedWings".

julianoes: Ok, seems good to me.

JonasVautherin: Actually, this advice came from Kenton Varda, [here on StackOverflow](https://stackoverflow.com/questions/31768665/can-i-define-a-grpc-call-with-a-null-request-or-response):

> ... we as developers are really bad at guessing what we might want in the future. So I recommend being safe by always defining custom params and results types for every method, even if they are empty.

I think I will also add custom responses...

julianoes:
@JonasVautherin is a custom error per command really practical? I would keep it simple stupid.

JonasVautherin:
@julianoes: It is a custom Response, not a custom error. Still following Kenton's advice to allow for evolving the protocol.

I also asked Thomas Vosst, and he agrees on the *Request and *Response pattern. Moreover, if you look at Google examples (for instance, check [in the styleguide](https://developers.google.com/protocol-buffers/docs/style)), they seem to use this pattern everywhere (at least on the examples I've read).

I agree that it makes the protocol definition a bit more verbose, but it doesn't add complexity as most messages are trivial. And I don't think this can hurt performances.

What do you guys think?

JonasVautherin:
Quick examples:

* Say we decide to add an `ActionResult` to `GetMaximumSpeedResult` at some point
* For the telemetry, right now we send `Position`s in the `Home` stream. What if at some point we decided that we actually want something different for `Home`, or to add a sanity field saying "valid" or "too far"?

action/action.proto
shakthi-prashanth-m: OK.

julianoes:
@JonasVautherin right, ok, agreed.

PR #6, opened 2018-01-24
define telemetry protocol

### LINKED ISSUE #4: Create proto file for telemetry plugin

JonasVautherin:
Fixes #4.

### DIFF
--- a/telemetry/telemetry.proto
+++ b/telemetry/telemetry.proto
@@ -6,15 +6,90 @@ option java_package = "io.dronecore.telemetry";
 option java_outer_classname = "TelemetryProto";
 
 service TelemetryService {
-    rpc SubscribePosition(SubscribePositionRequest) returns(stream Position) {}
+    rpc SubscribePosition(SubscribePositionRequest) returns(stream PositionResponse) {}
+    rpc SubscribeHome(SubscribeHomeRequest) returns(stream HomeResponse) {}
+    rpc SubscribeInAir(SubscribeInAirRequest) returns(stream InAirResponse) {}
+    rpc SubscribeArmed(SubscribeArmedRequest) returns(stream ArmedResponse) {}
+    rpc SubscribeAttitudeQuaternion(SubscribeAttitudeQuaternionRequest) returns(stream AttitudeQuaternionResponse) {}
+    rpc SubscribeAttitudeEuler(SubscribeAttitudeEulerRequest) returns(stream AttitudeEulerResponse) {}
+    rpc SubscribeCameraAttitudeQuaternion(SubscribeCameraAttitudeQuaternionRequest) returns(stream CameraAttitudeQuaternionResponse) {}
+    rpc SubscribeCameraAttitudeEuler(SubscribeCameraAttitudeEulerRequest) returns(stream CameraAttitudeEulerResponse) {}
+    rpc SubscribeGroundSpeedNED(SubscribeGroundSpeedNEDRequest) returns(stream GroundSpeedNEDResponse) {}
+    rpc SubscribeGPSInfo(SubscribeGPSInfoRequest) returns(stream GPSInfoResponse) {}
+    rpc SubscribeBattery(SubscribeBatteryRequest) returns(stream BatteryResponse) {}
+    rpc SubscribeFlightMode(SubscribeFlightModeRequest) returns(stream FlightModeResponse) {}
+    rpc SubscribeHealth(SubscribeHealthRequest) returns(stream HealthResponse) {}
+    rpc SubscribeRCStatus(SubscribeRCStatusRequest) returns(stream RCStatusResponse) {}
 }
 
-message UUID {
-    uint64 value = 1;
+message SubscribePositionRequest {}
+message PositionResponse {
+    Position position = 1;
 }
 
-message SubscribePositionRequest {
-    UUID uuid = 1;
+message SubscribeHomeRequest {}
+message HomeResponse {
+    Position home = 1;
+}
+
+message SubscribeInAirRequest {}
+message InAirResponse {
+    bool is_in_air = 1;
+}
+
+message SubscribeArmedRequest {}
+message ArmedResponse {
+   bool is_armed = 1;
+}
+
+message SubscribeAttitudeQuaternionRequest {}
+message AttitudeQuaternionResponse {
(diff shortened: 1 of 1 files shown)


telemetry/telemetry.proto
```diff
@@ -6,15 +6,90 @@ option java_package = "io.dronecore.telemetry";
...
+    rpc SubscribeAttitudeEuler(SubscribeAttitudeEulerRequest) returns(stream AttitudeEulerResponse) {}
+    rpc SubscribeCameraAttitudeQuaternion(SubscribeCameraAttitudeQuaternionRequest) returns(stream CameraAttitudeQuaternionResponse) {}
+    rpc SubscribeCameraAttitudeEuler(SubscribeCameraAttitudeEulerRequest) returns(stream CameraAttitudeEulerResponse) {}
+    rpc SubscribeGroundSpeedNED(SubscribeGroundSpeedNEDRequest) returns(stream GroundSpeedNEDResponse) {}
+    rpc SubscribeGPSInfo(SubscribeGPSInfoRequest) returns(stream GPSInfoResponse) {}
+    rpc SubscribeBattery(SubscribeBatteryRequest) returns(stream BatteryResponse) {}
+    rpc SubscribeFlightMode(SubscribeFlightModeRequest) returns(stream FlightModeResponse) {}
+    rpc SubscribeHealth(SubscribeHealthRequest) returns(stream HealthResponse) {}
+    rpc SubscribeRCStatus(SubscribeRCStatusRequest) returns(stream RCStatusResponse) {}
 }
```
shakthi-prashanth-m: How do we handle `Unsubscribe()` ? ; we pass `nullptr` in case of  C++.

julianoes:
@JonasVautherin actually what's your opinion on?

`SubscribeRCStatus` vs `SubscribeRcStatus`

JonasVautherin:
@julianoes: I'm okay with both. I just copied "RCStatus" from telemetry_impl.h, but if you find it more readable I can change it to SubscribeRcStatus without a problem.

I see RC the same way as GPS, that we usually don't write as "Gps" (again, I don't have a very strong opinion on that). I have a stronger opinion on "Takeoff" because it is one word (see [Wikipedia](https://en.wikipedia.org/wiki/Takeoff)).

telemetry/telemetry.proto
JonasVautherin: That's a good question.

I see two ways:

1. Maybe we can have an `Unsubscribe*` equivalent for each stream, which would make the stream end. I am not sure how clean that is on the server side.
2. (Currently my preferred one) we let the client unsubscribe from the stream whenever it wants, in a pubsub manner.

I checked online on how to make the client unsubscribe from a server stream in gRPC, and [found this for Java](https://github.com/grpc/grpc-java/pull/3115). It allows the client to "cancel" a server stream. I therefore assumed that it should be possible to do that from other languages, but it's right that we will need to test it more thoroughly.

shakthi-prashanth-m: OK Cool.

telemetry/telemetry.proto
```diff
@@ -23,3 +98,68 @@ message Position {
...
+    int32 num_satellites = 1;
+    FixType fix_type = 2;
+}
+
+enum FixType {
+    NO_GPS = 0;
+    NO_FIX = 1;
+    FIX_2D = 2;
+    FIX_3D = 3;
+    FIX_DGPS = 4;
```
shakthi-prashanth-m: Missing **5: RTK float**, **6: RTK fixed**

telemetry/telemetry.proto
```diff
@@ -23,3 +98,68 @@ message Position {
...
+}
+
+message Battery {
+    float voltage = 1;
```
shakthi-prashanth-m: It should be `voltage_v` with units.

telemetry/telemetry.proto
```diff
@@ -23,3 +98,68 @@ message Position {
...
+    MISSION = 3;
+    RETURN_TO_LAUNCH = 4;
+    LAND = 5;
+    OFFBOARD = 6;
+    FOLLOW_ME = 7;
+    UNKNOWN = 1000;
+}
+
+message Health {
+    bool is_gyrometer_calibration_ok = 1;
```
shakthi-prashanth-m: Nice! `is_` is more readable!

telemetry/telemetry.proto
```diff
@@ -23,3 +98,68 @@ message Position {
...
+    bool is_accelerometer_calibration_ok = 2;
+    bool is_magnetometer_calibration_ok = 3;
+    bool is_level_calibration_ok = 4;
+    bool is_local_position_ok = 5;
+    bool is_global_position_ok = 6;
+    bool is_home_position_ok = 7;
+}
+
+message RCStatus {
+    bool was_available_once = 1;
```
shakthi-prashanth-m: This one's Nice too! Its more meaningful to say `was_available_once = true`,

telemetry/telemetry.proto
```diff
@@ -23,3 +98,68 @@ message Position {
...
+    float z = 4;
+}
+
+message EulerAngle {
+    float roll_deg = 1;
+    float pitch_deg = 2;
+    float yaw_deg = 3;
+}
+
+message SpeedNED {
```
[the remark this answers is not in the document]
JonasVautherin: Yeah but both `GroundSpeedNED` and `AirSpeedNED` are "NED" speeds, right? We have a `SpeedNED` in the `GroundSpeedNEDResponse`, but if we had an `AirSpeedNEDResponse`, we could reuse it, couldn't we?

telemetry/telemetry.proto
JonasVautherin: Oh, interesting! I copied a comment in the telemetry plugin. Adding that right away!

JonasVautherin: My bad, I misread the comment actually -_-

telemetry/telemetry.proto
JonasVautherin: Good catch!

telemetry/telemetry.proto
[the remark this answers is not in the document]
shakthi-prashanth-m: Yeah, true; we can reuse.

julianoes:
Ok let's leave it as is.

PR #7, opened 2018-01-24
define mission protocol

### LINKED ISSUE #5: Create proto file for mission plugin

JonasVautherin:
Fixes #5.

### DIFF
--- a/mission/mission.proto
+++ b/mission/mission.proto
@@ -6,25 +6,69 @@ option java_package = "io.dronecore.mission";
 option java_outer_classname = "MissionProto";
 
 service MissionService {
-    rpc SendMission(SendMissionRequest) returns(MissionResult) {}
-    rpc StartMission(StartMissionRequest) returns(MissionResult) {}
+    rpc UploadMission(UploadMissionRequest) returns(UploadMissionResponse) {}
+    rpc DownloadMission(DownloadMissionRequest) returns(DownloadMissionResponse) {}
+    rpc StartMission(StartMissionRequest) returns(StartMissionResponse) {}
+    rpc PauseMission(PauseMissionRequest) returns(PauseMissionResponse) {}
+    rpc SetCurrentMissionItemIndex(SetCurrentMissionItemIndexRequest) returns(SetCurrentMissionItemIndexResponse) {}
+    rpc GetCurrentMissionItemIndex(GetCurrentMissionItemIndexRequest) returns(GetCurrentMissionItemIndexResponse) {}
+    rpc GetMissionCount(GetMissionCountRequest) returns(GetMissionCountResponse) {}
+    rpc IsMissionFinished(IsMissionFinishedRequest) returns(IsMissionFinishedResponse) {}
+    rpc SubscribeMissionProgress(SubscribeMissionProgressRequest) returns(stream MissionProgressResponse) {}
 }
 
-message UUID {
-    uint64 value = 1;
+message UploadMissionRequest {
+    Mission mission = 1;
+}
+message UploadMissionResponse {
+    MissionResult mission_result = 1;
 }
 
-message SendMissionRequest {
-    UUID uuid = 1;
-    Mission mission = 2;
+message DownloadMissionRequest {}
+message DownloadMissionResponse {
+    Mission mission = 1;
 }
 
-message StartMissionRequest {
-    UUID uuid = 1;
+message StartMissionRequest {}
+message StartMissionResponse {
+    MissionResult mission_result = 1;
+}
+
+message PauseMissionRequest {}
+message PauseMissionResponse {
+    MissionResult mission_result = 1;
+}
+
+message SetCurrentMissionItemIndexRequest {
+    int32 index = 1;
+}
+message SetCurrentMissionItemIndexResponse {
+    MissionResult mission_result = 1;
+}
+
+message GetCurrentMissionItemIndexRequest {}
+message GetCurrentMissionItemIndexResponse {
+    int32 index = 1;
+}
+
+message GetMissionCountRequest {}
+message GetMissionCountResponse {
+    int32 count = 1;
+}
+
+message IsMissionFinishedRequest {}
+message IsMissionFinishedResponse {
+    bool is_finished = 1;
+}
+
+message SubscribeMissionProgressRequest {}
+message MissionProgressResponse {
+    int32 current_item_index = 1;
+    int32 mission_count = 2;
 }
 
 message Mission {
-    repeated MissionItem mission_items = 1;
+    repeated MissionItem mission_item = 1;
 }
 
 message MissionItem {
@@ -58,6 +102,8 @@ message MissionResult {
         BUSY = 3;
         TIMEOUT = 4;
         INVALID_ARGUMENT = 5;
-        UNKNOWN = 6;
+        UNSUPPORTED = 6;
+        NO_MISSION_AVAILABLE = 7;
+        UNKNOWN = 1000;
     }
 }


mission/mission.proto
```diff
@@ -6,25 +6,63 @@ option java_package = "io.dronecore.mission";
...
-    rpc SendMission(SendMissionRequest) returns(MissionResult) {}
-    rpc StartMission(StartMissionRequest) returns(MissionResult) {}
+    rpc UploadMission(UploadMissionRequest) returns(UploadMissionResponse) {}
+    rpc DownloadMission(DownloadMissionRequest) returns(DownloadMissionResponse) {}
+    rpc StartMission(StartMissionRequest) returns(StartMissionResponse) {}
+    rpc PauseMission(PauseMissionRequest) returns(PauseMissionResponse) {}
+    rpc SetCurrentMissionItemId(SetCurrentMissionItemIdRequest) returns(SetCurrentMissionItemIdResponse) {}
+    rpc GetCurrentMissionItemId(GetCurrentMissionItemIdRequest) returns(GetCurrentMissionItemIdResponse) {}
+    rpc GetMissionSize(GetMissionSizeRequest) returns(GetMissionSizeResponse) {}
+    rpc SubscribeMissionProgress(SubscribeMissionProgressRequest) returns(stream MissionProgressResponse) {}
```
Rjasuja: We can add ClearMission to delete all mission items at once

julianoes: We now have sync and async functionality about the current mission.
@JonasVautherin do you think we should expose both through gRPC or only one of the two. This applies generall, I guess.

shakthi-prashanth-m: It should be supported from DroneCore C++ first :)

Rjasuja: It is part of Mavlink message specs

mission/mission.proto
```diff
@@ -6,25 +6,63 @@ option java_package = "io.dronecore.mission";
...
 service MissionService {
-    rpc SendMission(SendMissionRequest) returns(MissionResult) {}
-    rpc StartMission(StartMissionRequest) returns(MissionResult) {}
+    rpc UploadMission(UploadMissionRequest) returns(UploadMissionResponse) {}
+    rpc DownloadMission(DownloadMissionRequest) returns(DownloadMissionResponse) {}
+    rpc StartMission(StartMissionRequest) returns(StartMissionResponse) {}
+    rpc PauseMission(PauseMissionRequest) returns(PauseMissionResponse) {}
+    rpc SetCurrentMissionItemId(SetCurrentMissionItemIdRequest) returns(SetCurrentMissionItemIdResponse) {}
+    rpc GetCurrentMissionItemId(GetCurrentMissionItemIdRequest) returns(GetCurrentMissionItemIdResponse) {}
+    rpc GetMissionSize(GetMissionSizeRequest) returns(GetMissionSizeResponse) {}
```
shakthi-prashanth-m:
```
message GetMissionCountRequest {}
message GetMissionCountResponse {
    int32 mission_count = 1;
}
rpc GetMissionCount(GetMissionCountRequest) returns(GetMissionCountResponse) {}
```
Isn't this intuitive ?

mission/mission.proto
```diff
@@ -6,25 +6,63 @@ option java_package = "io.dronecore.mission";
 option java_outer_classname = "MissionProto";
 
 service MissionService {
-    rpc SendMission(SendMissionRequest) returns(MissionResult) {}
-    rpc StartMission(StartMissionRequest) returns(MissionResult) {}
+    rpc UploadMission(UploadMissionRequest) returns(UploadMissionResponse) {}
+    rpc DownloadMission(DownloadMissionRequest) returns(DownloadMissionResponse) {}
+    rpc StartMission(StartMissionRequest) returns(StartMissionResponse) {}
+    rpc PauseMission(PauseMissionRequest) returns(PauseMissionResponse) {}
+    rpc SetCurrentMissionItemId(SetCurrentMissionItemIdRequest) returns(SetCurrentMissionItemIdResponse) {}
```
shakthi-prashanth-m:
```
rpc SetCurrentMissionItemIndex(SetCurrentMissionItemIndexRequest) returns(SetCurrentMissionItemIndexResponse) {}
```

```
message SetCurrentMissionItemIndexRequest {
    int32 item_index = 1;
}
 message SetCurrentMissionItemIndexResponse {
     MissionResult mission_result = 1;
 }
```
How about this ?

mission/mission.proto
shakthi-prashanth-m: We need one for [mission_finished()](https://github.com/dronecore/DroneCore/blob/904ba588f554562672e79ea9dcc9449096219d47/plugins/mission/mission.h#L125).
Probably,
```
message IsMissionDoneRequest {}
message IsMissionDoneResponse {
     bool is_done = 1;
 }

rpc IsMissionDone(IsMissionDoneRequest) returns(IsMissionDoneResponse) {}
```

mission/mission.proto
```diff
@@ -6,25 +6,63 @@ option java_package = "io.dronecore.mission";
...
+    rpc GetCurrentMissionItemId(GetCurrentMissionItemIdRequest) returns(GetCurrentMissionItemIdResponse) {}
```
shakthi-prashanth-m:
```
message GetCurrentMissionItemIndexRequest {}
message GetCurrentMissionItemIndexResponse {
    int32 item_index = 1;
}

rpc GetCurrentMissionItemIndex(GetCurrentMissionItemIndexRequest) returns(GetCurrentMissionItemIndexResponse) {}
```

mission/mission.proto
```diff
@@ -6,25 +6,63 @@ option java_package = "io.dronecore.mission";
...
+message PauseMissionResponse {
+    MissionResult mission_result = 1;
+}
+
+message SetCurrentMissionItemIdRequest {
+    int32 item_id = 1;
+}
+message SetCurrentMissionItemIdResponse {
+    MissionResult mission_result = 1;
+}
```
shakthi-prashanth-m:
```
message SetCurrentMissionItemIndexRequest {
    int32 item_index = 1;
}
message SetCurrentMissionItemIndexResponse {
    MissionResult mission_result = 1;
}
```

mission/mission.proto
shakthi-prashanth-m: True, but we need a hook in DroneCore C++ interface.

JonasVautherin: Good points. Nothing prevents us from extending the protocol later. So as soon as we have a "ClearMission" implementation on the DroneCore libs, we can add it to the protocol :-).

mission/mission.proto
JonasVautherin: Fine for me, I'll change it.

mission/mission.proto
JonasVautherin: Not sure about this one. But it's right that `MissionSize` might sound like "the distance covered by the mission" or something like this.

But I don't know if `MissionCount` is intuitive to me (but if it is for the others, I'm fine with it).

Or maybe `MissionItemsCount` or `MissionItemsNumber`?

I'd like to get more opinions on this one :-).

shakthi-prashanth-m:
@JonasVautherin I've one question. If you see these [examples](https://github.com/grpc/grpc/blob/master/examples/protos/route_guide.proto), param names aren't extra-qualified. In case of message definition say,
```
message SetCurrentMissionItemIndexRequest {
    int32 index = 1;
}
```
is fine instead of saying `item_idex`, as it obviously refers to item,'s index.
What do you think ?

mission/mission.proto
shakthi-prashanth-m: Same as that of [MISSION_COUNT](http://mavlink.org/messages/common#MISSION_COUNT) in MAVLink.

shakthi-prashanth-m: `MissionItemCount`  is also fine I think.

JonasVautherin: If that's the MAVLink message, then I guess it makes sense to use it!

mission/mission.proto
JonasVautherin: `IsMissionFinished`, for consistency?

JonasVautherin:
Fair point. I was not sure as it is a `*Request` (and in the route example, they really have structures), but I probably overthought this. I changed it for `index`, as you suggest.

And I applied the same for `GetMissionCountResponse` (`mission_count` -> `count`).

shakthi-prashanth-m:
@JonasVautherin OK That makes sense. Thanks.

mission/mission.proto
shakthi-prashanth-m: Yeah, fine.

mission/mission.proto
```diff
@@ -6,25 +6,69 @@ option java_package = "io.dronecore.mission";
...
+message IsMissionFinishedRequest {}
+message IsMissionFinishedResponse {
+    bool is_finished = 1;
 }
 
-message StartMissionRequest {
-    UUID uuid = 1;
+message SubscribeMissionProgressRequest {}
+message MissionProgressResponse {
+    int32 current_item_id = 1;
```
shakthi-prashanth-m: `current_item_id` -> `current_item_index`

PR #12, opened 2018-01-25
define discovery protocol

### LINKED ISSUE #1: Create a "discovery" proto interface
Clients need an entrypoint to the backend. Because even the `core` component might end up managing only one vehicle (i.e. there will be multiple `core` components running at the same time), we need some way to get a list of the `core` components (e.g. with their "device uuid" and "gRPC server address:port").

This should be defined as a protobuf interface before it can be implemented on the backend side.

## Example 

A client wants to control a vehicle B through its plugins B1, B2, B3. It goes through the following:

1. The client contacts the entrypoint and gets a list of running `core` servers [A, B, C], each of them having its own <IP:port> pair (it could be the same one for all of them, but that is an implementation detail for the client).
2. The client selects the <IP:port> of `core` B and contacts it to fetch the list of plugins for vehicle B.
3. `core` B sends the list [B1, B2, B3] to the client, each component having its own <IP:port> pair. Again, whether the components share the <IP:port> pair or not is an implementation detail for the client.
4. The client can now interact with the drone through `core` and its corresponding plugins.

JonasVautherin:
This is trying to define the client interface for multi-devices (feel free to suggest other ways if you find this unintuitive!).

* A client fetches the list of known devices through a "discovery" module.
* If a client doesn't know any discovery module, it can start one on its own.
* The discovery module listens for both dronecore instances and mavlink instances (for the latter, the client might want to start a dronecore instance).
* For each discovered device, the module gives a UUID and an address, and says if it is a dronecore instance or not (in which case it is a mavlink instance).
* When a client knows the address of the dronecore instance it wants to use, it can lists the corresponding plugins and get, for each of them, the address of the corresponding grpc server.

In the first iteration, we will not have any discovery module (so we will always assume the client knows the address of the dronecore instance) and the plugins will run behind the same grpc server as the core. But at least the API will be ready for the future.

Fixes #1.

### DIFF
--- a/core/core.proto
+++ b/core/core.proto
@@ -6,15 +6,16 @@ option java_package = "io.dronecore.core";
 option java_outer_classname = "CoreProto";
 
 service CoreService {
-    rpc SubscribeDevices(SubscribeDevicesRequest) returns(stream Device) {}
+    rpc ListRunningPlugins(ListRunningPluginsRequest) returns(ListRunningPluginsResponse) {}
 }
 
-message UUID {
-    uint64 value = 1;
+message ListRunningPluginsRequest {}
+message ListRunningPluginsResponse {
+    repeated PluginInfo = 1;
 }
 
-message SubscribeDevicesRequest {}
-
-message Device {
-    UUID uuid = 1;
+message PluginInfo {
+    string name = 1;
+    string address = 2;
+    int32 port = 3;
 }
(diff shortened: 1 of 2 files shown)


JonasVautherin:
Not `discovery.proto`, but that was to give a feeling of how I would see it, and it doesn't hurt to have it already defined but not used. I can remove it for now if you want.

`core.proto`, I think we need it, as it defines the interface on the frontend side. But the `dronecore_server` will have the plugins running in-process and "hard-coded", so it can already return a list of `PluginInfo`. It's just that at first they will all have the same address and port as the core.

julianoes:
Ok, makes sense.

PR #20, opened 2018-05-28
Add camera status and settings

### LINKED ISSUE #14: Add camera proto

JonasVautherin:
* Add missing SubscribeCameraStatus
* Update the settings interface with SubscribeCurrentSetting, SubscribePossibleSettings, SetSetting.

The idea is that when you SubscribePossibleSettings, you receive all the settings and the options to which you can set them, together with a description (of the setting and its options). When you SubscribeCurrentSettings and SetSettings, it is assumed you know the settings and options by id already.

Fixes #14.


camera/camera.proto
```diff
@@ -159,16 +167,16 @@ message VideoStreamSettings {
...
         IN_PROGRESS = 1;
     }
 
     VideoStreamSettings video_stream_settings = 1;
-    Status status = 2;
+    VideoStreamStatus video_stream_status = 2;
 }
 
-message Status {
+message CameraStatus {
```
julianoes: Isn't everything namespaced by camera or do we need this?

PR #25, opened 2018-06-27
Parse return type of methods

JonasVautherin:
Parsing the return type of methods (single, observable) and passing it to the templates.

For generation purpose, `*Response` messages can now have a `Result` type (e.g. `MissionResult`, `ActionResult`, ...) and maximum one parameter (which may be a tuple).

### DIFF
--- a/pb_plugins/dcsdkgen/struct.py
+++ b/pb_plugins/dcsdkgen/struct.py
@@ -2,6 +2,7 @@
 
 
 from .utils import (extract_string_type,
+                    is_request,
                     is_response,
                     is_struct)
 
@@ -36,7 +37,17 @@ class Struct(object):
         return _structs
 
     @staticmethod
-    def collect_responses(package, structs, template_env):
+    def collect_requests(package, structs):
+        _requests = {}
+
+        for struct in structs:
+            if is_request(struct):
+                _requests[struct.name] = struct
+
+        return _requests
+
+    @staticmethod
+    def collect_responses(package, structs):
         _responses = {}
 
         for struct in structs:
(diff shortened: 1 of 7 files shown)


pb_plugins/dcsdkgen/utils.py
```diff
@@ -55,6 +60,12 @@ def extract_string_type(field):
         return "UNKNOWN_TYPE"
 
 
+def filter_out_result(fields):
+    """ Filters out the result fields (".*Result$") """
+    for field in fields:
+        if not extract_string_type(field).endswith("Result"): yield field
```
xvzf: - Styling; yield in new line
- Add e.g. `is_result(..)` which contains `extract_string_type(field).endswith("Result")`

protos/mission/mission.proto
```diff
@@ -92,6 +91,11 @@ message MissionItem {
     }
 }
 
+message MissionProgress {
```
xvzf: > Responses cannot have more than 1 return parameter

This is relating, right?

pb_plugins/dcsdkgen/struct.py
```diff
@@ -36,7 +37,17 @@ def collect_structs(package, structs, template_env):
         return _structs
 
     @staticmethod
-    def collect_responses(package, structs, template_env):
+    def collect_requests(package, structs):
```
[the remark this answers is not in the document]
JonasVautherin: Good point. It would, probably.

protos/mission/mission.proto
JonasVautherin: Yes. An observable returns a stream of objects of the same type or an error.

If you look at the swift implementation, actually, the two return parameters of this response were actually packaged inside a tuple anyway.

xvzf: Okay, I that's what I was thinking!

xvzf:
Sort of related to this PR:
I would propose on moving the template files to the actual SDK repos and just keep `dcsdkgen` in here.

PR #29, opened 2018-07-30
[WIP] Improve python templates

xvzf:


pb_plugins/dcsdkgen/utils.py
```diff
@@ -80,13 +80,17 @@ def decapitalize(s):
...
+
+
 def jinja_indent(_in_str, level):
     """ Indentation helper for the jinja2 templates """
 
     _in_str = str(_in_str)
 
     return "\n".join(
-        ["" if len(line) == 0 else
+        ["" if not line else
```
JonasVautherin: :D

pb_plugins/dcsdkgen/methods.py
```diff
@@ -141,10 +175,11 @@ def __init__(
             requests,
             responses):
         super().__init__(plugin_name, package, pb_method, requests, responses)
-        self._name = decapitalize(pb_method.name) + "Observable"
-        self._capitalized_name = pb_method.name + "Observable"
+        self._name = remove_subscribe(decapitalize(pb_method.name))
```
JonasVautherin: That changes the behavior, right? It means that the Swift template now needs to add "Observable". Is that correct?

xvzf: Yes, it makes the name more generalized. The SWIFT templates have to be adjusted in some places, I'll do that asap!

PR #30, opened 2018-08-16
Handle complex return types

JonasVautherin:
Check if the return type is a primitive type or not. If not, the message/enum needs to be translated.

@xvzf

### DIFF
--- a/pb_plugins/dcsdkgen/utils.py
+++ b/pb_plugins/dcsdkgen/utils.py
@@ -62,6 +62,10 @@ def extract_string_type(field):
     else:
         return "UNKNOWN_TYPE"
 
+def is_primitive_type(field):
+    """ Check if the field type is primitive (e.g. bool,
+    float) or not (e.g message, enum) """
+    return (not field.type in { 11, 14 })
 
 def filter_out_result(fields):
     """ Filters out the result fields (".*Result$") """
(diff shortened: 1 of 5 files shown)


pb_plugins/dcsdkgen/utils.py
```diff
@@ -62,6 +62,10 @@ def extract_string_type(field):
     else:
         return "UNKNOWN_TYPE"
 
+def is_primitive_type(field):
+    """ Check if the field type is primitive (e.g. bool,
+    float) or not (e.g message, enum) """
+    return (not field.type in { 11, 14 })
```
xvzf: `(...)` is optional ;)

pb_plugins/templates/swift/method_single.j2
```diff
@@ -3,8 +3,14 @@ public func {{ name }}() -> Single<{{ return_type }}> {
...
-            let {{ name }}Response = try self.service.{{ name.lower() }}({{ name }}Request)
-            single(.success({{ name }}Response.{{ return_name }}))
+            let response = try self.service.{{ name.lower() }}({{ name }}Request)
+            {% if is_return_type_primitive -%}
+               let {{ return_name }} = response.{{ return_name }}
+            {% else -%}
+                let {{ return_name }} = {{ return_type }}.translateFromRPC(response.{{ return_name }})
+            {%- endif %}
+
+            single(.success({{ return_name }}))
```
xvzf: Nice!

pb_plugins/dcsdkgen/utils.py
xvzf: How do you  get to 11 and 14?

JonasVautherin: https://github.com/grpc/grpc/blob/master/third_party/nanopb/generator/proto/google/protobuf/descriptor.proto#L128

PR #34, opened 2018-08-21
Add grpc plugins

JonasVautherin:
The templates are not completely finished, but the parsing is complete.

I think we should merge that and continue working on it directly in `master`.

@xvzf @julianoes: what do you think?


xvzf:
Wait until tomorrow! 
I'll go over the code again :)

xvzf:
Went over the code. If we merge #35 we could merge this to master and continue working there on the templates!

PR #45, opened 2018-09-19
Ignore struct template if it does not exist

JonasVautherin:
When generating the backend code, I don't actually need to generate the structs (but I need to collect them, so `struct.py` is needed but not `struct.j2`).

I don't know if that's the best way to ignore it. Do you guys have an opinion?


pb_plugins/dcsdkgen/struct.py
```diff
@@ -11,8 +11,13 @@ class Struct(object):
     """ Struct """
 
     def __init__(self, package, template_env, pb_struct):
+        try:
+            self._template = template_env.get_template("struct.j2")
+        except:
```
xvzf: We should __NEVER__ do that! We might miss some exceptions.

JonasVautherin: Right, sorry :sweat_smile:. Is it better like this? Or should I avoid the `try...except` syntax altogether?

PR #48, opened 2018-09-20
Added name class

### LINKED ISSUE #46: Consistently capitalize names in jinja2
Different languages have different naming conventions (upper/lower camel case, snake case, ...). We need a way for the jinja2 templates to use the naming convention specific to their need. Note that our proto files use upper camel case.

Following discussions with @xvzf, a proposition would be to have class representing a name (/string) and providing conversions. For instance:

`name = CaseName("CalibrateGimbalGyro")` would provide:

* name.upper_camel_case
* name.lower_camel_case
* name.snake_case

`name` would then be passed to the jinja2 template and could be used there (e.g. `{{ name.snake_case }}()` would call `calibrate_gimbal_gyro()`).

@julianoes @xvzf

xvzf:
Untested. What do you think @JonasVautherin 

Fixes #46.

### DIFF
--- a/pb_plugins/dcsdkgen/enum.py
+++ b/pb_plugins/dcsdkgen/enum.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-from .name import Name
+from .name import NameParser
 
 
 class Enum(object):
@@ -8,7 +8,7 @@ class Enum(object):
     def __init__(self, package, template_env, pb_enum):
         self._package = package
         self._template = template_env.get_template("enum.j2")
-        self._name = Name(pb_enum.name)
+        self._name = NameParser(pb_enum.name)
         self._values = []
 
         for value in pb_enum.value:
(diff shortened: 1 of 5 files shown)


JonasVautherin:
Note, if you say "Fixes #46" in the PR description, [waffle](https://waffle.io/Dronecode/DronecodeSDK) closes the issue when the PR is merged, which I find convenient :-).

pb_plugins/dcsdkgen/enum.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from .name import Name
```
JonasVautherin: Do we want to call the class `Name`? Like `plugin_name` will be a `Name` as well (which is why I tried "CaseName" in the issue, which is not ideal).

pb_plugins/dcsdkgen/name.py
```diff
@@ -0,0 +1,23 @@
+import re
+
+
+class Name:
+    """ Implements https://github.com/Dronecode/DronecodeSDK-Proto/issues/46
```
JonasVautherin: I would remove that :-)

PR #49, opened 2018-09-24
Update code for using NameParser

JonasVautherin:
Trying to consistently use `NameParser` in the templates.

Also renamed `name.py` into `name_parser.py`


pb_plugins/dcsdkgen/autogen.py
```diff
@@ -42,7 +43,8 @@ def generate_reactive(request):
                                              responses,
                                              template_env)
 
-            out_file = File(plugin_name,
+            out_file = File(proto_file.package,
```
xvzf: We should fix on one parameter sequence, With `Struct.collect_structs`, `plugin_name`goes in front of `proto_file.package` and here it's swapped around

pb_plugins/dcsdkgen/name_parser.py
```diff
@@ -18,4 +22,8 @@ def lower_camel_case(self):
     def snake_case(self):
         return re.sub('(?<!^)(?=[A-Z])',
                       '_',
-                      self._name).lower()
+                      self._name)
+
+    @property
+    def lower_snake_case(self):
+        return self.snake_case.lower()
```
xvzf: Nice!

pb_plugins/dcsdkgen/struct.py
```diff
@@ -5,42 +5,40 @@
...
-            self._template = template_env.get_template("struct.j2")
-        except TemplateNotFound:
-            # We don't always generate code related to structs (but still
-            # collect them)
-            pass
-
+    def __init__(self, plugin_name, package, template_env, pb_struct):
+        self._plugin_name = NameParser(plugin_name)
+        self._package = NameParser(package)
+        self._template = template_env.get_template("struct.j2")
```
xvzf: Was it intended to remove the `TemplateNotFound` exception?

pb_plugins/dcsdkgen/struct.py
```diff
@@ -5,42 +5,40 @@
...
         self._name = NameParser(pb_struct.name)
         self._fields = []
-        self._rpc_type = (package.title().replace(".", "_")
-                          + "_"
-                          + self._name)
 
         for field in pb_struct.field:
-            self._fields.append((field.json_name, extract_string_type(field)))
+            self._fields.append(
+                (NameParser(
```
xvzf: Please keep the parameter of `NameParser` in the same line, I had to look over it three times to get it!

pb_plugins/dcsdkgen/utils.py
```diff
@@ -34,10 +34,9 @@ def is_response(struct):
 
 def is_struct(struct):
     """ Check if the message is a data structure or an rpc
-    request/response/result """
+    request/response"""
     return (not struct.name.endswith("Request") and
-            not struct.name.endswith("Response") and
-            not struct.name.endswith("Result"))
+            not struct.name.endswith("Response"))
```
xvzf: Can't e.g. an ActionResult be a struct? I am thinking of this one: https://github.com/Dronecode/DronecodeSDK-Proto/blob/74b6477739fac183b14fbc8df64fc9f71a538c06/protos/action/action.proto#L92

pb_plugins/dcsdkgen/utils.py
```diff
@@ -100,9 +99,6 @@ def get_template_env(_searchpath):
         searchpath=_searchpath))
 
     # Register some functions we need to access in the template
-    _template_env.globals.update(
-        letter_case_to_delimiter=letter_case_to_delimiter,
-        indent=jinja_indent,
-        remove_subscribe=remove_subscribe)
+    _template_env.globals.update(indent=jinja_indent)
```
xvzf: +1

We should also remove code for `letter_case_to_delimiter` and `remove_subscribe`, right?

pb_plugins/dcsdkgen/struct.py
JonasVautherin: Yes, actually I need the `struct` template in the backend, too :sweat_smile:

pb_plugins/dcsdkgen/struct.py
JonasVautherin: autopep8 is doing that :-/. I will check if there is a way to configure it.

pb_plugins/dcsdkgen/utils.py
JonasVautherin: `ActionResult` is a struct, but not `*Request` and `*Response`. I was ignoring `Result`s before, but I think I was wrong.

pb_plugins/dcsdkgen/utils.py
JonasVautherin: Oh, right, I'll check that!

pb_plugins/dcsdkgen/utils.py
xvzf: Actually I think I just missed the `not`. Should be fine!

pb_plugins/dcsdkgen/type_info.py
```diff
@@ -0,0 +1,50 @@
...
+
+class TypeInfo:
+    _default_types = {
+        1: "double",
+        2: "float",
+        4: "uint64_t",
+        5: "int",
+        8: "bool",
+        9: "std::string",
+        13: "uint32_t"
```
julianoes: I wonder where we need `uint32_t`.

JonasVautherin: Not sure. I just translated the types I knew (i.e. the ones we use), but it would make sense here to have all the proto types (it's a plugin for the proto idl after all).

julianoes: I would like to constrain the types that we need a bit rather than having all of them.

JonasVautherin: Well we don't "have them". The plugin knows how to translate a `uint32_t`, that's all. What we do with that is defined by our proto files. In the same way, we don't fork protobuf to not "have `uint32_t` there. We just don't use it in our proto files, that's it.

PR #50, opened 2018-09-25
Add TypeInfo to manage types and their conversions

### LINKED ISSUE #47: Provide type conversions between languages
Related to #46, but slightly different.

Different languages name the same primitive types differently (be it because of naming conventions, or bigger differences like "bool vs boolean"). We need a way to convert the type to each of our generated languages.

Following a discussion with @xvzf, a proposition would be to set the mapping between types in some configuration file. The reference could be C primitive types. The following example illustrates such a configuration file for Java:

```
int -> int
double -> double
bool -> boolean
string -> String
```

This file would be read and used by the parser to generate the correct type for jinja2. This time, the conversion would be done on the parser side (e.g. one would pass `return_type = TypeConverter.format("string")` to jinja2, and a java template would hence have "String" for `return_type`).

@julianoes @xvzf

JonasVautherin:
This is a proposition for #47. It looks for a file called `type_conversions` in the templates folder, and if existing, uses it for converting c++ types into whatever is defined in the file.

The file is expected to be a json object where keys are the c++ types and values are their conversions.

Fixes #47

### DIFF
--- a/pb_plugins/dcsdkgen/type_info.py
+++ b/pb_plugins/dcsdkgen/type_info.py
@@ -1,0 +1,50 @@
+import json
+from os import environ
+
+
+class TypeInfo:
+    _default_types = {
+        1: "double",
+        2: "float",
+        4: "uint64_t",
+        5: "int",
+        8: "bool",
+        9: "std::string",
+        13: "uint32_t"
+    }
+
+    try:
+        _conversion_dict_path = environ.get(
+            "TEMPLATE_PATH", "./") + "/type_conversions"
+        with open(_conversion_dict_path, "r") as handle:
+            _conversion_dict = json.loads(handle.read())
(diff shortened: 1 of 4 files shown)


pb_plugins/dcsdkgen/type_info.py
```diff
@@ -0,0 +1,50 @@
...
+        1: "double",
+        2: "float",
+        4: "uint64_t",
+        5: "int",
+        8: "bool",
+        9: "std::string",
+        13: "uint32_t"
+    }
+
+    try:
```
JonasVautherin: This is probably ugly, but I did not know how to do it otherwise. Open to propositions :-).

xvzf: Can't think of anything that's more elegant!

PR #53, opened 2018-10-08
snake_case should be upper_snake_case for consistency

JonasVautherin:
That's a detail, but it was bothering me.


JonasVautherin:
:laughing:

PR #55, opened 2018-10-11
Test type info

JonasVautherin:
Loading the file statically is not nice for unit testing. But loading the file for each instantiation doesn't sound right either. So I made a factory for that, that loads the file only once.

### DIFF
--- a/pb_plugins/dcsdkgen/type_info.py
+++ b/pb_plugins/dcsdkgen/type_info.py
@@ -2,6 +2,24 @@ import json
 from os import environ
 
 
+class TypeInfoFactory:
+
+    def __init__(self):
+        self._conversion_dict = self._load_conversions_dict()
+
+    def create(self, field):
+        return TypeInfo(field, self._conversion_dict)
+
+    def _load_conversions_dict(self):
+        try:
+            _template_path = environ.get("TEMPLATE_PATH", "./")
+            _conversion_dict_path = f"{_template_path}/type_conversions"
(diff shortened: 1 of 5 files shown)


pb_plugins/dcsdkgen/type_info.py
```diff
@@ -2,6 +2,24 @@
...
 
+class TypeInfoFactory:
+
+    def __init__(self):
+        self._conversion_dict = self.__load_conversions_dict()
+
+    def create(self, field):
+        return TypeInfo(field, self._conversion_dict)
+
+    def __load_conversions_dict(self):
```
xvzf: `__` is usually an python internal function name, I'd suggest to go with just one `_`!

pb_plugins/dcsdkgen/type_info.py
```diff
@@ -2,6 +2,24 @@
...
+        try:
+            _conversion_dict_path = environ.get(
```
xvzf: Could be more readable when using an formatted String (`f"this is a text with {variable_name_or_function}"`

pb_plugins/dcsdkgen/type_info.py
JonasVautherin: `_` is private, and `__` internal then?

xvzf: Exactly! There is no such thing as private in python but it is the common way to do it over here :)

JonasVautherin:
@xvzf Should be better now! And done!

PR #57, opened 2018-11-13
A message name cannot be identical to the plugin name

JonasVautherin:
Because depending of the language, a plugin becomes a class (`<Plugin>`), and a message becomes a class as well (`<Plugin>`), resulting in a conflict.

### DIFF
--- a/protos/mission/mission.proto
+++ b/protos/mission/mission.proto
@@ -20,7 +20,7 @@ service MissionService {
 }
 
 message UploadMissionRequest {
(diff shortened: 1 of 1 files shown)


JonasVautherin:
Well actually I think I prefer the later solution. And we'll possibly have some weird `mission_items.mission_items` calls in the generated code.

julianoes:
Ha, just too late, now you have to make a new PR.

PR #58, opened 2018-11-13
add support for repeated types in TypeInfo

JonasVautherin:
The conversion dict file now can have a `repeated` entry that defines the `prefix` and `suffix` for a repeated type.

For instance, the default is:

    { "repeated", { "prefix": "std::vector<", "suffix": ">" } }

Meaning that a `repeated double` will become `std::vector<double>`.

### DIFF
--- a/pb_plugins/dcsdkgen/type_info.py
+++ b/pb_plugins/dcsdkgen/type_info.py
@@ -30,7 +30,8 @@ class TypeInfo:
         8: "bool",
         9: "std::string",
         12: "std::byte",
-        13: "uint32_t"
+        13: "uint32_t",
+        "repeated": {"prefix": "std::vector<", "suffix": ">"}
     }
 
     def __init__(self, field, conversion_dict):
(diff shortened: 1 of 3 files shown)


pb_plugins/dcsdkgen/type_info.py
```diff
@@ -61,3 +77,9 @@ def is_primitive(self):
         """ Check if the field type is primitive (e.g. bool,
         float) or not (e.g message, enum) """
         return self._field.type not in {11, 14}
+
+    @property
+    def is_repeated(self):
+        """ Check if the field is a repeated type (in which
+        case it is converted to a list) """
+        return self._field.label == 3
```
julianoes: Shouldn't this number 3 be in some define list?

xvzf: +1, I also think this should be defined somewhere

JonasVautherin: :+1: 

I'll check how I can do that!

PR #63, opened 2019-01-06
pb_plugins: improve name_parser inputs handling

JonasVautherin:
It was assuming upper camel case before, but it actually happens that the input arrives as lower_snake_case.

Also, uppercase should keep underscores (e.g. `TakePhoto` should become `TAKE_PHOTO` and not `TAKEPHOTO`, which doesn't really make sense).

Finally, lowercase is apparently superfluous.

### DIFF
--- a/pb_plugins/dcsdkgen/name_parser.py
+++ b/pb_plugins/dcsdkgen/name_parser.py
@@ -4,15 +4,27 @@ import re
 class NameParser:
 
     def __init__(self, name):
-        self._name = name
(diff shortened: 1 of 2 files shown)


JonasVautherin:
> How do you run these tests?

The quick way is to run `test_name_parser.py`. There is a better way running the tests in different versions of python through Docker, and I should put that on a CI, but haven't had the time yet.

xvzf:
You can also run `tox` on the local machine without Docker in order to test pultiple python versions!

PR #65, opened 2019-01-22
pb_plugins: add support for initialisms

JonasVautherin:
Some languages find that "Url" is much more difficult to understand than "URL", and that it is worth breaking the cONsiStENCy of CamelCase by keeping initialisms capitalized. And because it is not so easy to have a list of such initialisms, there are only 4 of them in swift-protobuf (nice, isn't it? :sweat_smile:). 

We have to deal with it, because otherwise our generated code doesn't compile with code generated by swift-protobuf. Therefore, I added a list of initialisms in `NameParser`, so that we can break our consistency, too, for those languages.

Notes:
* `NameParser` takes a list of initialisms without a default, because I want to force the use of `NameParserFactory`. 
* The way I treat the initialisms below (with long `map` lines) is not super elegant, but I did not find a better way (I was pissed off, because [look how elegant](https://github.com/Dronecode/DronecodeSDK-Proto/pull/65/files#diff-57beef92659b7dc537130208ff200706R65) it can be without initialisms... :sweat_smile:). Open for suggestions!

### DIFF
--- a/pb_plugins/dcsdkgen/autogen_file.py
+++ b/pb_plugins/dcsdkgen/autogen_file.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-from .name_parser import NameParser
+from .utils import name_parser_factory
 
 
 class File(object):
@@ -14,8 +14,8 @@ class File(object):
             structs,
             methods,
             has_result):
-        self._package = NameParser(package)
-        self._plugin_name = NameParser(plugin_name)
+        self._package = name_parser_factory.create(package)
+        self._plugin_name = name_parser_factory.create(plugin_name)
         self._template = template_env.get_template("file.j2")
         self._enums = enums
         self._structs = structs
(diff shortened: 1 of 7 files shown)


julianoes:
Honestly I disagree and I vote for consistency. :confused:

JonasVautherin:
Sorry, I was not super clear in my description above. I __hate__ that choice. Especially because having a list of 4 initialisms means that even the treatment of initialisms is not consistent, because there are _many_ of them that are not in this non-exhaustive list of size 4.

However, we don't have a choice: the code generated by swift-protobuf treats those 4 initialisms differently, and we happen to have `file_url` (and `id`, I believe) in our proto files. So our generated code needs to call `fileURL`.

I thought about exposing `fileUrl` on our side, and calling the rpc `fileURL` in the implementation, but that would make the whole `NameParser` more complicated because it would have to know about API/implementation, and that's even worse.

PR #67, opened 2019-01-31
mission: add camera_photo_interval parameter to MissionItem

JonasVautherin:
For some reason it was not there :sweat_smile:.


julianoes:
Funny.

PR #70, opened 2019-02-15
Add template_path as a parameter to protoc plugin

JonasVautherin:
Our protoc plugin now takes `template_path` from the protoc options. The goal is to not have to use the `TEMPLATE_PATH` environment variable. I find it cleaner, and used it in the gradle integration.

The `TEMPLATE_PATH` environment variable can still be used, but should
probably be deprecated some time soon. Or do you guys like having both possibilities? I would vote for aiming at having one way to do a thing :-).


JonasVautherin:
Merging for now, I can improve it later if @xvzf has comments ;-).

PR #71, opened 2019-02-20
dcsdkgen: add parent_type to type info

JonasVautherin:
In some languages (like Python), nested enums need to know their parent name. This adds it to type_info.

This is assuming only one level of nesting. It could probably be generalized, but we don't have such an example yet.

It also assumes (for now) that the package name has 3 levels (e.g. "org.dronecode.sdk"), but I believe I could solve that with the generalized nesting handling.

Required to fix [this issue](https://github.com/Dronecode/DronecodeSDK-Python/issues/47).

### DIFF
--- a/pb_plugins/dcsdkgen/type_info.py
+++ b/pb_plugins/dcsdkgen/type_info.py
@@ -72,6 +72,24 @@ class TypeInfo:
             return self._default_types[type_id]
 
     @property
+    def parent_type(self):
+        """ Extracts parent type. This assumes that nesting is not deeper
+        than 1 level.
+
+        Note that it also assumes that the package name has 3 words. It
+        will start with a '.', giving something like:
+        '.org.dronecode.sdk.ActionResult.Result'."""
+        if self.is_primitive:
+            return None
+
+        type_tree = self._field.type_name.split(".")
+
+        if len(type_tree) <= 5:
+            return None
+
+        return str(type_tree[-2])
+
+    @property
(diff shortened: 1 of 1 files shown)


pb_plugins/dcsdkgen/type_info.py
```diff
@@ -71,6 +71,20 @@ def inner_name(self):
...
+    @property
+    def parent_type(self):
+        """ Extracts parent type. This assumes that nesting is not deeper
+        than 1 level """
+        if self.is_primitive:
+            return None
+
+        type_tree = self._field.type_name.split(".")
+
+        if len(type_tree) <= 5:
```
julianoes: Why 5?

JonasVautherin: Because "org.dronecode.sdk" becomes ".org.dronecode.sdk", and splitting that on `.`'s gives an empty string in the beginning.

I don't really want to hardcode that, but I am not sure how this works and for now it's really based on my observations (they don't really have extensive documentation regarding writing custom protoc plugins). I will improve it as soon as I have more data to understand better where the type comes from :sweat_smile:.

julianoes: Oh, I mean if you can put something like that in the comment that might help the reader.

JonasVautherin: Like that? I added information to the function comment.

PR #73, opened 2019-02-21
Various fixes

JonasVautherin:
* Fix build for multiple proto files (previously it was returning after the first file)
* Improve error handling when protoc is run with the `file_ext` option missing
* Output java files in the right directory, according to java conventions (i.e. use the package)

### DIFF
--- a/pb_plugins/dcsdkgen/autogen.py
+++ b/pb_plugins/dcsdkgen/autogen.py
@@ -22,6 +22,8 @@ class AutoGen(object):
         type_info_factory.set_template_path(params["template_path"])
         template_env = get_template_env(params["template_path"])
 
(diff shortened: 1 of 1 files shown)


pb_plugins/dcsdkgen/autogen.py
```diff
@@ -57,14 +59,12 @@ def generate_reactive(request):
...
 
-            _codegen_response = plugin_pb2.CodeGeneratorResponse()
-
             # Fill response
             f = _codegen_response.file.add()
             f.name = f"{plugin_name}.{params['file_ext']}"
             f.content = str(out_file)
 
-            return _codegen_response
+        return _codegen_response
```
julianoes: Ha!

PR #74, opened 2019-02-22
Add wrappers for cancel calibration, mission download, and mission upload.

douglaswsilva:
- Add cancel calibration wrapper.
- Add cancel mission download wrapper.
- Add cancel mission upload wrapper.

### DIFF
--- a/protos/calibration/calibration.proto
+++ b/protos/calibration/calibration.proto
@@ -38,9 +38,7 @@ message CalibrateGimbalAccelerometerResponse {
 }
 
 message CancelCalibrationRequest {}
-message CancelCalibrationResponse {
-     CalibrationResult calibration_result = 1;
-}
+message CancelCalibrationResponse {}
 
 message CalibrationResult {
     enum Result {
--- a/protos/calibration/calibration.proto
+++ b/protos/calibration/calibration.proto
@@ -10,7 +10,6 @@ service CalibrationService {
     rpc SubscribeCalibrateAccelerometer(SubscribeCalibrateAccelerometerRequest) returns(stream CalibrateAccelerometerResponse) {}
     rpc SubscribeCalibrateMagnetometer(SubscribeCalibrateMagnetometerRequest) returns(stream CalibrateMagnetometerResponse) {}
     rpc SubscribeCalibrateGimbalAccelerometer(SubscribeCalibrateGimbalAccelerometerRequest) returns(stream CalibrateGimbalAccelerometerResponse) {}
-    rpc Cancel(CancelCalibrationRequest) returns(CancelCalibrationResponse) {}
 }
 
 message SubscribeCalibrateGyroRequest {}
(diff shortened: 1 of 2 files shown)


protos/calibration/calibration.proto
```diff
@@ -10,6 +10,7 @@ service CalibrationService {
     rpc SubscribeCalibrateAccelerometer(SubscribeCalibrateAccelerometerRequest) returns(stream CalibrateAccelerometerResponse) {}
     rpc SubscribeCalibrateMagnetometer(SubscribeCalibrateMagnetometerRequest) returns(stream CalibrateMagnetometerResponse) {}
     rpc SubscribeCalibrateGimbalAccelerometer(SubscribeCalibrateGimbalAccelerometerRequest) returns(stream CalibrateGimbalAccelerometerResponse) {}
+    rpc Cancel(CancelCalibrationRequest) returns(CancelCalibrationResponse) {}
```
JonasVautherin: Does that work :thinking: ? I believe that my templates assume that it's the same prefix everywhere:

`Cancel`, `CancelRequest`, `CancelResponse`, or
`CancelCalibration`, `CancelCalibrationRequest`, `CancelCalibrationResponse`

Then I think I would vote for `Cancel`, because it will then become `calibration.cancel()` instead of `calibration.cancelCalibration()`.

protos/mission/mission.proto
```diff
@@ -7,7 +7,9 @@ option java_outer_classname = "MissionProto";
 
 service MissionService {
     rpc UploadMission(UploadMissionRequest) returns(UploadMissionResponse) {}
+    rpc UploadMissionCancel(UploadMissionCancelRequest) returns(UploadMissionCancelResponse) {}
```
JonasVautherin: Wouldn't it be clearer to call that `CancelMissionUpload`? Wondering :slightly_smiling_face:.

protos/mission/mission.proto
```diff
@@ -7,7 +7,9 @@ option java_outer_classname = "MissionProto";
...
     rpc DownloadMission(DownloadMissionRequest) returns(DownloadMissionResponse) {}
+    rpc DownloadMissionCancel(DownloadMissionCancelRequest) returns(DownloadMissionCancelResponse) {}
```
JonasVautherin: ... and `CancelMissionDownload` here, then.

protos/calibration/calibration.proto
douglaswsilva: Got it. I did have to modify something in the generated SDK. I'll see what the consensus is in the other channel about the naming convention and change it here. :)

protos/mission/mission.proto
douglaswsilva: Yup, agree. I've been just following the C++ SDK naming. So the discussion it falls in the same category as the cancel calibration thing. I can Do the updates here first and you guys change the SDK later?

JonasVautherin: I think so. But we need @julianoes to agree on that, too :-).

PR #76, opened 2019-02-26
Add params

douglaswsilva:
- Add params


JonasVautherin:
> I think option 2 is what I would go for.

But then that should get reflected in the C++ API, right?

PR #77, opened 2019-02-28
Add param proto

douglaswsilva:
- Add protofile for params.


douglaswsilva:
Yeah, I was not sure how to name it. But if there's a clear distinction between "params raw" and some other type of "params" I can change the name to ParamsRaw. Other question is whether to call it "Param" singular (like we do for Mission, Action, etc). or "Params" plural.

julianoes:
Yer I was thinking about singular vs plural as well.

JonasVautherin:
I don't have a super strong opinion, but I would vote for `Param` instead of `Params`. Other than that, it looks good to me! Thanks a lot!

PR #80, opened 2019-03-28
mission: add missing result

julianoes:
Anything else that needs to be added for this?

### DIFF
--- a/protos/mission/mission.proto
+++ b/protos/mission/mission.proto
@@ -119,5 +119,6 @@ message MissionResult {
         FAILED_TO_OPEN_QGC_PLAN = 9;
         FAILED_TO_PARSE_QGC_PLAN = 10;
         UNSUPPORTED_MISSION_CMD = 11;
+        TRANSFER_CANCELLED = 12;
     }
 }


protos/mission/mission.proto
```diff
@@ -119,5 +119,6 @@ message MissionResult {
         FAILED_TO_OPEN_QGC_PLAN = 9;
         FAILED_TO_PARSE_QGC_PLAN = 10;
         UNSUPPORTED_MISSION_CMD = 11;
+        CANCELLED = 12;
```
JonasVautherin: Wait, does it mean "UPLOAD_CANCELLED"?

julianoes: Or download cancelled.

JonasVautherin: But not `MISSION_CANCELLED`, right? What about `TRANSFER_CANCELLED` then?

julianoes: Right that would be better but you'd have to change it here first:
https://github.com/Dronecode/DronecodeSDK/blob/develop/plugins/mission/include/plugins/mission/mission.h#L54

JonasVautherin: The gRPC server can make the translation until we update the C++ API. Given that the goal will be eventually to generate the C++ API from these proto files, right?

JonasVautherin:
Should we do something about that? I still vote for `TRANSFER_CANCELLED`, and I volunteer to adapt the C++ part when the headers get generated :D

julianoes:
Ok @JonasVautherin whatever you suggest.

JonasVautherin:
@julianoes may I ask you to merge, as a sign for "approval"? It's too easy for me to edit your PR and then approve it :sweat_smile:.

PR #81, opened 2019-04-22
Add Offboard Proto

cswkim:
One note, I see `NO_SETPOINT_SET` as part of the `Result` enum in the documentation for `develop` [here](https://sdk.dronecode.org/en/api_reference/classdronecode__sdk_1_1_offboard.html#classdronecode__sdk_1_1_offboard_1a1f515eed9d23c450254233ea87131c09), should this be explicitly returned [here](https://github.com/Dronecode/DronecodeSDK/blob/a2b9db92f83e53bcce75c611f6e9b473f4341562/plugins/offboard/offboard.cpp#L51-L70)?

### DIFF
--- a/protos/offboard/offboard.proto
+++ b/protos/offboard/offboard.proto
@@ -1,0 +1,95 @@
+syntax = "proto3";
+
+package dronecode_sdk.rpc.offboard;
+
+option java_package = "io.dronecode_sdk.offboard";
+option java_outer_classname = "OffboardProto";
+
+service OffboardService {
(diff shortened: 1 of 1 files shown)


julianoes:
Thanks for the proto definition. @JonasVautherin please review.

And good find, we should fix the `result_str()`.

JonasVautherin:
@cswkim: would you mind opening an issue in DronecodeSDK for the `result_str`?

cswkim:
@JonasVautherin I'll open an issue for the `result_str` and will reach out on slack about the gRPC backend support.

PR #82, opened 2019-04-25
Add mavlink messages support.

douglaswsilva:
Hey guys, I'm implementing the reading mavlink messages from the SDK and was wondering if this is the best way to do it. Please let me know what you think. 

Note that I tried to use `DEBUG` in the enum, but the compiler complained, so I'm using `DEBUS = 7;` for now, let me know what would be a good alternative word.

### DIFF
--- a/protos/telemetry/telemetry.proto
+++ b/protos/telemetry/telemetry.proto
@@ -20,6 +20,7 @@ service TelemetryService {
     rpc SubscribeFlightMode(SubscribeFlightModeRequest) returns(stream FlightModeResponse) {}
     rpc SubscribeHealth(SubscribeHealthRequest) returns(stream HealthResponse) {}
     rpc SubscribeRcStatus(SubscribeRcStatusRequest) returns(stream RcStatusResponse) {}
+    rpc SubscribeMavMessage(SubscribeMavMessageRequest) returns(stream MavMessageResponse) {}
 }
 
 message SubscribePositionRequest {}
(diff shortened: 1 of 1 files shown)


protos/telemetry/telemetry.proto
```diff
@@ -20,6 +20,7 @@ service TelemetryService {
     rpc SubscribeFlightMode(SubscribeFlightModeRequest) returns(stream FlightModeResponse) {}
     rpc SubscribeHealth(SubscribeHealthRequest) returns(stream HealthResponse) {}
     rpc SubscribeRcStatus(SubscribeRcStatusRequest) returns(stream RcStatusResponse) {}
+    rpc SubscribeMavMessage(SubscribeMavMessageRequest) returns(stream MavMessageResponse) {}
```
julianoes: I wouldn't call it `MavMessage`, what about `StatusText`?

protos/telemetry/telemetry.proto
```diff
@@ -165,3 +171,20 @@ message RcStatus {
...
+    enum Type {
+        EMERGENCY = 0;
+        ALERT = 1;
+        CRITICAL = 2;
+        ERROR = 3;
+        WARNING = 4;
+        NOTICE = 5;
+        INFO = 6;
+        DEBUS = 7;
+        UNKNOWN = 8;
```
julianoes: The thing is that PX4 only ever uses `INFO`, `WARNING`, and `CRITICAL`, so we might as well drop the others.

JonasVautherin: Good to know. And we can bring the others back if needed later.

protos/telemetry/telemetry.proto
douglaswsilva: Coll, will change the name. :)

protos/telemetry/telemetry.proto
douglaswsilva: Ah, as Jonas said, good to know. I'll drop them.

PR #84, opened 2019-06-05
camera: remove set video stream settings

julianoes:
This setting is no longer available.

See:
https://github.com/Dronecode/DronecodeSDK/pull/756
https://github.com/mavlink/mavlink/pull/1053


julianoes:
> (it is only used by SetVideoStreamSettings).

No, it's also used in `VideoStreamInfo`.

PR #87, opened 2019-06-18
added SetActuatorControl method with corresponding messages

irsdkv:
This is a corresponding PR to one made in DronecodeSDK [here](https://github.com/mavlink/MAVSDK/pull/771)
Python primitive types translates: https://github.com/mavlink/MAVSDK-Python/pull/76

### DIFF
--- a/protos/offboard/offboard.proto
+++ b/protos/offboard/offboard.proto
@@ -10,6 +10,7 @@ service OffboardService {
     rpc Stop(StopRequest) returns(StopResponse) {}
     rpc IsActive(IsActiveRequest) returns(IsActiveResponse) {}
     rpc SetAttitude(SetAttitudeRequest) returns(SetAttitudeResponse) {}
+    rpc SetActuatorControl(SetActuatorControlRequest) returns(SetActuatorControlResponse) {}
     rpc SetAttitudeRate(SetAttitudeRateRequest) returns(SetAttitudeRateResponse) {}
     rpc SetPositionNed(SetPositionNedRequest) returns(SetPositionNedResponse) {}
     rpc SetVelocityBody(SetVelocityBodyRequest) returns(SetVelocityBodyResponse) {}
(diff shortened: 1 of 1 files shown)


protos/offboard/offboard.proto
```diff
@@ -63,6 +69,18 @@ message Attitude {
...
+message ActuatorControl {
+  enum ActuatorGroup {
+    FLIGHT_CONTROL = 0;
+    FLIGHT_CONTROL_VTOL = 1;
+    GIMBAL = 2;
+    MANUAL_PASSTHROUGH = 3;
+  }
+
+  ActuatorGroup actuator_group = 1;
+  repeated float actuator_values = 2 [packed=true];
```
JonasVautherin: I believe `[packed=true]` is not necessary with [proto3](https://developers.google.com/protocol-buffers/docs/proto3):

> In proto3, repeated fields of scalar numeric types use packed encoding by default.

JonasVautherin:
Thanks a lot for the PRs @irsdkv! Sorry for the delay, we were busy at the PX4 DevSummit last week :-).

I made a small comment, but other than that it is looking good to me. I will however wait for a review from @julianoes, who is more knowledgeable than me on that matter :-). He is on holiday this week, so it will have to wait until next Monday. Let me know if it is urgent for you and we can look for another solution :+1:.

irsdkv:
> Thanks a lot for the PRs @irsdkv! Sorry for the delay, we were busy at the PX4 DevSummit last week :-).
> 
> I made a small comment, but other than that it is looking good to me. I will however wait for a review from @julianoes, who is more knowledgeable than me on that matter :-). He is on holiday this week, so it will have to wait until next Monday. Let me know if it is urgent for you and we can look for another solution .

Thank for your remark, it is true :)

Sure, it is not necessary for us to do it today. let's wait until next week :)

PR #88, opened 2019-07-02
Add direct control without specifying group names

irsdkv:
This is a corresponding PR to one made in DronecodeSDK [here](https://github.com/mavlink/MAVSDK/pull/782)
Python primitive types translates: https://github.com/mavlink/MAVSDK-Python/pull/76


JonasVautherin:
Thanks a lot @irsdkv!

PR #89, opened 2019-07-16
Add geofence proto file

coderkalyan:
This is, at the very least, a draft for a proto file for adding geofence capabilities to MAVSDK-Python and others. Note that this is my first time working with .proto, so feel free to let me know if something is incorrect!


protos/geofence/geofence.proto
```diff
@@ -0,0 +1,49 @@
+syntax = "proto3";
+
+package mavsdk.rpc.geofence;
+
+option java_package = "io.mavlink.mavsdk.geofence";
+option java_outer_classname = "GeofenceProto";
+
+service GeofenceService {
+    rpc SendGeofence(SendGeofenceRequest) returns(SendGeofenceRespose) {}
```
JonasVautherin:
```suggestion
    rpc SendGeofence(SendGeofenceRequest) returns(SendGeofenceResponse) {}
```

protos/geofence/geofence.proto
```diff
@@ -0,0 +1,49 @@
...
+
+    repeated Point points = 1;
+    Type type = 2;
+}
+    
+message SendGeofenceRequest {
+    repeated Polygon polygons = 1;
+}
+
+message SendGeofenceRespose {
```
JonasVautherin:
```suggestion
message SendGeofenceResponse {
```

protos/geofence/geofence.proto
```diff
@@ -0,0 +1,49 @@
...
+    enum Type {
+        INCLUSION = 0;
+        EXCLUSION = 1;
+    }
+
+    repeated Point points = 1;
+    Type type = 2;
+}
+    
+message SendGeofenceRequest {
```
julianoes: Can we name this consistent with the Mission proto? So we also use `upload` and `download? @JonasVautherin what do you think?

JonasVautherin: Good catch! I agree!

coderkalyan: @julianoes So to clarify, you would like me to change `SendGeofenceRequest` to `UploadGeofenceRequest`? Also, I more or less copied the names from the C++ core - would you like me to look into changing that too?

julianoes: Ah sorry, I can see how that is confusing. Yes, let's do `Upload`.

coderkalyan: I updated the proto file. Sorry for a couple of messy commits, I ran into merge conflicts since I forgot that some typos were fixed online.

coderkalyan:
@JonasVautherin @julianoes I just realized that since Point was a submessage of Polygon, bindings weren't being generated properly for the Point class. I am not sure if this is a bug or not. Anyway, I have moved Point to the top level, so Point is now accessed (in python) as `mavsdk.Point`, not `mavsdk.Polygon.Point`. Please let me know if this is incorrect.

protos/geofence/geofence.proto
```diff
@@ -0,0 +1,49 @@
+syntax = "proto3";
+
+package mavsdk.rpc.geofence;
+
+option java_package = "io.mavlink.mavsdk.geofence";
```
JonasVautherin:
```suggestion
option java_package = "io.mavsdk.geofence";
```

JonasVautherin:
> I am not sure if this is a bug or not. Anyway, I have moved Point to the top level

Hmm there may be some work to do in the generator for nested types. This said, IMO it is fine to have `Point` at the top level.

coderkalyan:
@JonasVautherin @julianoes I have updated the java package, I think this PR is ready to merge? (Your call :) )

PR #93, opened 2019-07-23
Add gimbal mode support.

douglaswsilva:
Suggestion for initial idea of how set gimbal mode will look like.

For now it is working this way, but you guys might want to wait for: https://github.com/mavlink/mavlink/pull/1174 

Let's open a discussion here and see what to do.

### DIFF
--- a/protos/gimbal/gimbal.proto
+++ b/protos/gimbal/gimbal.proto
@@ -26,8 +26,8 @@ message SetModeResponse {
 }
 
 enum GimbalMode {
-    FOLLOW = 0;
-    LOCK = 1;
+    YAW_FOLLOW = 0;
+    YAW_LOCK = 1;
 }
 
 message GimbalResult {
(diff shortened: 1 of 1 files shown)


protos/gimbal/gimbal.proto
```diff
@@ -17,6 +18,18 @@ message SetPitchAndYawResponse {
...
+message SetModeRequest {
+    GimbalMode gimbal_mode = 1;
+}
+message SetModeResponse {
+    GimbalResult gimbal_result = 1;
+}
+
+enum GimbalMode {
+    FOLLOW = 0;
+    LOCK = 1;
```
julianoes: Please specify that this refers to YAW, so `YAW_FOLLOW` I'd say.

JonasVautherin: Is that strictly reserved to yaw, on the gimbal side? Couldn't it follow with pitch as well?

julianoes: `YAW_FOLLOW` means that the gimbal follows the autopilot yaw, so it always looks forward (although with a little bit of damping so the picture is smoother).
`YAW_LOCK` means that the gimbal locks onto a specific heading no matter what direction the drone faces.

Pitch is basically always in `LOCK` mode for multicopters because you don't want it to be influenced by the vehicle pitch.

PR #94, opened 2019-07-24
fix spelling: NED -> Ned

JonasVautherin:
@avdwerff: That's the reason for your build error.


avdwerff:
> @avdwerff: That's the reason for your build error.

Ok, so once approved and merged I will try again.

JonasVautherin:
> Ok, so once approved and merged I will try again.

@avdwerff: well, we need to propagate that to MAVSDK-Swift and to `mavsdk_server` (the latter is necessary only if you plan to use offboard). I'll mention you on those PRs.

PR #97, opened 2019-08-08
Added telemetry requests.

irsdkv:
Added:
 
- AttitudeAngularSpeed: current angular speed from [ATTITUDE_QUATERNION](https://mavlink.io/en/messages/common.html#ATTITUDE_QUATERNION) message.

-   ActuatorControlTarget: current vehicle [attitude and body angular rates](https://mavlink.io/en/messages/common.html#SET_ACTUATOR_CONTROL_TARGET).
    
-  ActuatorOutputStatus: current raw values of the [actuator outputs](https://mavlink.io/en/messages/common.html#ACTUATOR_OUTPUT_STATUS).

Related PR in backend: https://github.com/mavlink/MAVSDK/pull/832

### DIFF
--- a/protos/telemetry/telemetry.proto
+++ b/protos/telemetry/telemetry.proto
@@ -12,6 +12,7 @@ service TelemetryService {
     rpc SubscribeArmed(SubscribeArmedRequest) returns(stream ArmedResponse) {}
     rpc SubscribeAttitudeQuaternion(SubscribeAttitudeQuaternionRequest) returns(stream AttitudeQuaternionResponse) {}
     rpc SubscribeAttitudeEuler(SubscribeAttitudeEulerRequest) returns(stream AttitudeEulerResponse) {}
+    rpc SubscribeAttitudeAngularSpeed(SubscribeAttitudeAngularSpeedRequest) returns(stream AttitudeAngularSpeedResponse) {}
     rpc SubscribeCameraAttitudeQuaternion(SubscribeCameraAttitudeQuaternionRequest) returns(stream CameraAttitudeQuaternionResponse) {}
     rpc SubscribeCameraAttitudeEuler(SubscribeCameraAttitudeEulerRequest) returns(stream CameraAttitudeEulerResponse) {}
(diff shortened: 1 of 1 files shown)


protos/telemetry/telemetry.proto
```diff
@@ -118,6 +136,12 @@ message EulerAngle {
     float yaw_deg = 3;
 }
 
+message AngularSpeed {
+    float rollspeed = 1;
+    float pitchspeed = 2;
+    float yawspeed = 3;
+}
```
julianoes: I'm confused here. What is this for? Is it for this...

https://github.com/mavlink/MAVSDK/blob/935988bbf398cb4eb7c274b698e137073a8c9434/src/plugins/telemetry/include/plugins/telemetry/telemetry.h#L134-L143

irsdkv: This is for current angular speed from ATTITUDE_QUATERNION message:
https://github.com/mavlink/MAVSDK/pull/832/files#diff-5d3c38fd06bf641a3131f56ff3b57f05R117

julianoes: Hm, and that's not the same as `AngularVelocityNED` I suppose but rather in body frame? Could we get the naming closer to `AngularVelocityNED`?

irsdkv: Yes, this is from [this msg](https://mavlink.io/en/messages/common.html#ATTITUDE_QUATERNION), in body frame.

I think that naming really could be closer.
How about 
```
message AngularVelocityBF {
    float roll_rad_s = 1;
    float pitch_rad_s = 2;
    float yaw_rad_s = 3;
}
```

julianoes:
```
message AngularVelocityBody {
    float roll_rad_s = 1;
    float pitch_rad_s = 2;
    float yaw_rad_s = 3;
}
```

There are others called `...Body` so that would be consistent.

irsdkv: Ok. Will do it

irsdkv:
My bad. Done :-)

PR #98, opened 2019-08-17
Parse documentation

### LINKED ISSUE #91: Autogenerate documentation
Our `protoc-gen-dcsdk` protoc plugin already parses the proto files to feed them to our language templates in order to autogenerate language bindings in different languages. As an example, the MAVSDK-Swift [templates](https://github.com/mavlink/MAVSDK-Swift/tree/master/templates) end up in the MAVSDK-Swift [generated source files](https://github.com/mavlink/MAVSDK-Swift/tree/master/Sources/MAVSDK-Swift/Generated).

Ideally, we would like to add documentation (doxygen-like) inside this generated code, e.g. to document [the Action class](https://github.com/mavlink/MAVSDK-Swift/blob/master/Sources/MAVSDK-Swift/Generated/Action.swift#L5) or [the arm() completable](https://github.com/mavlink/MAVSDK-Swift/blob/master/Sources/MAVSDK-Swift/Generated/Action.swift#L156).

We would want this documentation to come respectively from [here](https://github.com/mavlink/MAVSDK-Proto/blob/master/protos/action/action.proto#L8) and [here](https://github.com/mavlink/MAVSDK-Proto/blob/master/protos/action/action.proto#L9), so that we write it once for all languages.

It is possible to parse comments in proto files, as is proven by [protoc-gen-doc](https://github.com/pseudomuto/protoc-gen-doc), a protoc plugin written in go that does just that. However, we want to go one step further: instead of generating documentation for the proto files, we want to feed that documentation into our templates so that it ends up in our generated source code. [This commit](https://github.com/grpc/grpc-java/commit/7e8b504e3f25c28f934b8a5ab6b1215c25aed5b1#diff-4e543a2f8a0243d4e0fed21737cdbf18) seems to be parsing documentation as well, in java.

I don't know how to parse documentation out of a proto file (but protoc-gen-doc seems to do that nicely), and I don't know how we should structure that documentation in the proto files in such a way that it can be used to generate language-specific documentation through templates (e.g. javadoc in Java, etc...).

The first step would be to make a proof of concept going in that direction, any contribution is welcome!

JonasVautherin:
This is adding the (admittedly) quite cryptic `docs.py` file, that tries to make sense of the comments in the proto files (based on documentation [here](https://github.com/protocolbuffers/protobuf/blob/a08b03d4c00a5793b88b494f672513f6ad46a681/src/google/protobuf/descriptor.proto)).

I believe this brings support for all the kind of documentation we currently need, including nested enums (one level only) and nested messages (which we currently don't use).

Resolves #91.

### DIFF
--- a/pb_plugins/protoc_gen_dcsdk/docs.py
+++ b/pb_plugins/protoc_gen_dcsdk/docs.py
@@ -1,0 +1,97 @@
+# -*- coding: utf-8 -*-
+
+
+class Docs():
+    """
+    Docs
+
+    Based on https://github.com/protocolbuffers/protobuf/blob/a08b03d4c00a5793b88b494f672513f6ad46a681/src/google/protobuf/descriptor.proto
+    """
+
+    @staticmethod
+    def collect_docs(source_code_info):
+
+        docs = {}
+
+        docs['methods'] = []
+        docs['structs'] = []
+        docs['enums'] = []
+
+        for location in source_code_info.location:
+            if len(location.path) > 0:
+                path = location.path
+
+                if path[0] == 1:  # name
+                    pass
+                elif path[0] == 2:  # package
+                    pass
+                elif path[0] == 4:  # message
+                    struct_id = path[1]
+
+                    if len(path) == 2:  # Top-level location of this message
+                        docs['structs'].append(
+                            {'description': location.leading_comments, 'params': []})
+                    # A field of this message
+                    elif len(path) == 4 and path[2] == 2:
+                        param_id = path[3]
+                        docs['structs'][struct_id]['params'].append(
+                            location.trailing_comments)
+                    elif path[2] == 3:  # A nested message
+                        nested_struct_id = path[3]
+
+                        if len(
+                                path) == 4:  # Top-level location of this nested message
+                            docs['structs'][struct_id]['structs'] = []
+                            docs['structs'][struct_id]['structs'].append(
+                                {'description': location.leading_comments, 'params': []})
+                        # A field of this nested message
+                        elif len(path) == 6 and path[4] == 2:
+                            docs['structs'][struct_id]['structs'][nested_struct_id]['params'].append(
+                                location.trailing_comments)
+                    elif path[2] == 4:  # A nested enum
+                        nested_enum_id = path[3]
+
+                        if len(path) == 4:  # Top-level location of this nested enum
+                            docs['structs'][struct_id]['enums'] = []
+                            docs['structs'][struct_id]['enums'].append(
+                                {'description': location.leading_comments, 'params': []})
+                        # A value of this nested message
+                        elif len(path) == 6 and path[4] == 2:
+                            docs['structs'][struct_id]['enums'][nested_enum_id]['params'].append(
+                                location.trailing_comments)
+
+                elif path[0] == 5:  # enum
+                    enum_id = path[1]
+
+                    if len(path) == 2:  # Top-level location of this enum
+                        docs['enums'].append(
+                            {'description': location.leading_comments, 'params': []})
+                    # A value of this message
+                    elif len(path) == 4 and path[2] == 2:
+                        param_id = path[3]
+                        docs['enums'][enum_id]['params'].append(
+                            location.trailing_comments)
+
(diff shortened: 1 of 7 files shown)


pb_plugins/protoc_gen_dcsdk/docs.py
```diff
@@ -0,0 +1,97 @@
+# -*- coding: utf-8 -*-
+
+
+class Docs(object):
```
julianoes: I think you don't need the `object` if it's only used with Python 3.

pb_plugins/protoc_gen_dcsdk/docs.py
```diff
@@ -0,0 +1,97 @@
...
+        for location in source_code_info.location:
+            if len(location.path) > 0:
+                path = location.path
+
+                if path[0] == 1:  # name
+                    pass
+                elif path[0] == 2:  # package
+                    pass
+                elif path[0] == 4:  # message
+                    struct_id = path[1]
```
julianoes: This would be more understandable if it branched out to methods.

pb_plugins/protoc_gen_dcsdk/docs.py
```diff
@@ -0,0 +1,97 @@
...
+                    # A field of this message
+                    elif len(path) == 4 and path[2] == 2:
+                        param_id = path[3]
+                        docs['structs'][struct_id]['params'].append(
+                            location.trailing_comments)
+                    elif path[2] == 3:  # A nested message
+                        nested_struct_id = path[3]
+
+                        if len(
+                                path) == 4:  # Top-level location of this nested message
```
julianoes: Weird newline.

pb_plugins/protoc_gen_dcsdk/enum.py
```diff
@@ -35,11 +42,19 @@ def collect_enums(
...
-            _enums[enum.name] = Enum(
-                plugin_name, package, template_env, enum, parent_struct)
+            _enums[enum.name] = Enum(plugin_name,
+                                     package,
+                                     template_env,
+                                     enum,
+                                     docs['enums'][enum_id] if docs else None,
+                                     parent_struct)
+
+            enum_id = enum_id + 1
```
julianoes: `enum_id += 1` ? :smile:

pb_plugins/protoc_gen_dcsdk/docs.py
JonasVautherin: Agreed, but that's `autopep8` :-/. Good thing is that the style is consistent everywhere, bad thing is that kind of things. I wouldn't know how to improve that, and I think it's not in the scope of this PR.

pb_plugins/protoc_gen_dcsdk/docs.py
JonasVautherin: Oh, didn't know that, I was blindly copying the other files :sweat_smile:. I'll change that everywhere then.

pb_plugins/protoc_gen_dcsdk/enum.py
JonasVautherin: Oh, I knew `enum_id++` was not Python, but I did not think about that :laughing:

pb_plugins/protoc_gen_dcsdk/docs.py
julianoes: Odd!

pb_plugins/protoc_gen_dcsdk/enum.py
julianoes: Hehe.

pb_plugins/protoc_gen_dcsdk/docs.py
JonasVautherin: I agree that there is some refactoring to be done here, but I'd be happier if we could get this in, so that I can move on with the templates, and come back later to that code.

The main reason being that I wrote this while generating Python documentation, and I'm not yet completely sure if that's good enough for all our languages or not. One we have it working for a few languages, I will know better how to structure this.

Does that make sense to you?

julianoes: Sure!

PR #99, opened 2019-08-19
add documentation to proto files

JonasVautherin:
Bring documentation from C++ to the proto files for a few plugins (action, calibration, camera, param, telemetry).

The other plugins will follow in a new PR soon.

@hamishwille: Those comments are the ones being parsed by #98 and used to generate language-specific documentation in the source files. I'll open a PR for the corresponding python templates as soon as those PRs are merged.

### DIFF
--- a/protos/action/action.proto
+++ b/protos/action/action.proto
@@ -5,21 +5,97 @@ package mavsdk.rpc.action;
 option java_package = "io.mavsdk.action";
 option java_outer_classname = "ActionProto";
 
+// Enable simple actions such as arming, taking off, and landing.
 service ActionService {
+    /*
+     * Send command to arm the drone.
+     *
+     * Arming a drone normally causes motors to spin at idle.
+     * Before arming take all safety precautions and stand clear of the drone!
+     */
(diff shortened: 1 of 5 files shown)


protos/action/action.proto
```diff
@@ -84,11 +160,11 @@ message SetTakeoffAltitudeResponse {
 message GetMaximumSpeedRequest {}
 message GetMaximumSpeedResponse {
     ActionResult action_result = 1;
-    float speed = 2;
+    float speed = 2; // Maximum speed (in metres/second)
```
julianoes: Not a change for this PR but we should decide on what we do about units!

JonasVautherin: What do you mean? In terms of how we document them?

julianoes: I think they should be part of the variable name (unless we can have units that reflect it).

JonasVautherin: Right. I don't think I have a strong opinion on that matter, but if we want to change, a good moment would be after we have the C++ auto-generation, and before we move to version 1.x :+1:.

PR #100, opened 2019-08-20
Revert "remove unused property"

coolkesword:
This reverts commit b34da699d08a1e8c9376e3feb25433d128080798.

Removal of parent_type method in type_info.py will lead to syntax error in generated python codes after executing: ./other/tools/run_protoc.sh. The parent_type method is needed in struct.j2 (line 46 & 49), so it cannot be removed. 

Thanks


coolkesword:
PS: The ./other/tools/run_protoc.sh script from the MAVSDK-Python repo, and struct.j2 is located in /other/templates/, so the pull request is made based on our MAVSDK-Python perspective.

PR #103, opened 2019-08-26
Logging raw streaming

irsdkv:
This is a draft to first Logging plugin API implementation.
Parser for ULog seems too complicated and redundant to implement it just now in C++ plugin for me.
Please let me know is that interface ok

### DIFF
--- a/protos/logging/logging.proto
+++ b/protos/logging/logging.proto
@@ -5,10 +5,11 @@ package mavsdk.rpc.logging;
 option java_package = "io.mavsdk.logging";
 option java_outer_classname = "LoggingProto";
 
+// Provide logging data.
 service LoggingService {
-    // Send command to start streaming logging data over MAVLink
+    // Start streaming logging data.
     rpc StartLogging(StartLoggingRequest) returns(StartLoggingResponse) {}
-    // Send command to stop streaming logging data over MAVLink
+    // Stop streaming logging data.
     rpc StopLogging(StopLoggingRequest) returns(StopLoggingResponse) {}
     // Subscribe to logging messages
     rpc SubscribeLoggingRaw(SubscribeLoggingRawRequest) returns(stream LoggingRawResponse) {}
@@ -29,6 +30,7 @@ message LoggingRawResponse {
     LoggingRaw logging_raw = 1; // A message containing logged data
 }
 
+// Raw logging data type
 message LoggingRaw {
     uint32 first_message_offset = 1; // Offset into data where first message starts
     bytes data = 2; // Logged data
--- a/protos/logging/logging.proto
+++ b/protos/logging/logging.proto
@@ -1,0 +1,52 @@
+syntax = "proto3";
+
+package mavsdk.rpc.logging;
+
+option java_package = "io.mavsdk.logging";
+option java_outer_classname = "LoggingProto";
+
+service LoggingService {
+    // Send command to start streaming logging data over MAVLink
+    rpc StartLogging(StartLoggingRequest) returns(StartLoggingResponse) {}
+    // Send command to stop streaming logging data over MAVLink
+    rpc StopLogging(StopLoggingRequest) returns(StopLoggingResponse) {}
+    // Subscribe to logging messages
+    rpc SubscribeLoggingRaw(SubscribeLoggingRawRequest) returns(stream LoggingRawResponse) {}
+}
+
+message StartLoggingRequest {}
+message StartLoggingResponse {
+    LoggingResult logging_result = 1;
+}
+
+message StopLoggingRequest {}
+message StopLoggingResponse {
+    LoggingResult logging_result = 1;
+}
+
+message SubscribeLoggingRawRequest {}
+message LoggingRawResponse {
+    LoggingRaw logging_raw = 1; // A message containing logged data
+}
+
+message LoggingRaw {
+    uint32 first_message_offset = 1; // Offset into data where first message starts
+    bytes data = 2; // Logged data
+}
+
+// Result type.
+message LoggingResult {
+    // Possible results returned for logging requests
+    enum Result {
+        SUCCESS = 0; // Request succeeded
+        NO_SYSTEM = 1; // No system connected
+        CONNECTION_ERROR = 2; // Connection error
+        BUSY = 3; // System busy
+        COMMAND_DENIED = 4; // Command denied
+        TIMEOUT = 5; // Timeout
+        UNKNOWN = 6; // Unknown error
+    }
+
+    Result result = 1; // Result enum value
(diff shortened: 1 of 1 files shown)


protos/logging/logging.proto
```diff
@@ -0,0 +1,52 @@
...
+    LoggingResult logging_result = 1;
+}
+
+message SubscribeLoggingRawRequest {}
+message SubscribeLoggingRawResponse {
+    LoggingRaw logging_raw = 1; // A message containing logged data
+}
+
+message LoggingRaw {
+    uint32 first_message_offset = 1; // Offset into data where first message starts
```
julianoes: So how exactly would this work? Why does it require this offset?

I'm asking because my opinion is that the API of MAVSDK should be as easy as possible, so I wonder how this would be used.

irsdkv: This is internal method to recover offset if data was lost
(https://mavlink.io/en/messages/common.html#LOGGING_DATA).
This will helpful for OTA processing, for example.

julianoes: Right, so this essentially exposes the actual MAVLink protocol? However, it internally would take care of retransmission of missed messages, right?

irsdkv: Retransmission is not available now in PX4 because of logging MAVLink stream require high bandwidth and CPU resources.

So we haven't way to request lost messages internal in MAVSDK.

I see only one way to avoid to send non-consistent data - parse [ULOG messages](https://dev.px4.io/v1.9.0/en/log/ulog_file_format.html) while receiving LOGGING_DATA MAVLink messages.
Without it backend should transmit full message's payload, I think.

julianoes: Aha. What about the headers, etc.? Are they just part of the first messages?

irsdkv: Yes, it is received in first 16 bytes on streaming start (I'm already checked that).

julianoes: ok

protos/logging/logging.proto
```diff
@@ -0,0 +1,52 @@
...
+message LoggingResult {
+    // Possible results returned for logging requests
+    enum Result {
+        SUCCESS = 0; // Request succeeded
+        NO_SYSTEM = 1; // No system connected
+        CONNECTION_ERROR = 2; // Connection error
+        BUSY = 3; // System busy
+        COMMAND_DENIED = 4; // Command denied
+        TIMEOUT = 5; // Timeout
+        UNKNOWN = 6; // Unknown error
```
JonasVautherin: Nit:pick: we usually have `UNKNOWN = 0`, and `SUCCESS = 1`. Would be more consistent :-).

protos/logging/logging.proto
```diff
@@ -0,0 +1,52 @@
...
+message StopLoggingResponse {
+    LoggingResult logging_result = 1;
+}
+
+message SubscribeLoggingRawRequest {}
+message LoggingRawResponse {
+    LoggingRaw logging_raw = 1; // A message containing logged data
+}
+
+message LoggingRaw {
```
JonasVautherin: Just a thought: what about `RawData` or `RawLoggingData`?

protos/logging/logging.proto
```diff
@@ -0,0 +1,52 @@
...
+message StopLoggingRequest {}
+message StopLoggingResponse {
+    LoggingResult logging_result = 1;
+}
+
+message SubscribeLoggingRawRequest {}
+message LoggingRawResponse {
+    LoggingRaw logging_raw = 1; // A message containing logged data
+}
+
```
JonasVautherin:
```suggestion

// Raw logging data type
```

protos/logging/logging.proto
```diff
@@ -0,0 +1,52 @@
+syntax = "proto3";
+
+package mavsdk.rpc.logging;
+
+option java_package = "io.mavsdk.logging";
+option java_outer_classname = "LoggingProto";
+
```
JonasVautherin:
```suggestion

// Provide logging data.
```

protos/logging/logging.proto
```diff
@@ -0,0 +1,52 @@
...
+service LoggingService {
+    // Send command to start streaming logging data over MAVLink
```
JonasVautherin:
```suggestion
    // Start streaming logging data.
```

protos/logging/logging.proto
```diff
@@ -0,0 +1,52 @@
...
+    rpc StartLogging(StartLoggingRequest) returns(StartLoggingResponse) {}
+    // Send command to stop streaming logging data over MAVLink
```
JonasVautherin:
```suggestion
    // Stop streaming logging data.
```

protos/logging/logging.proto
irsdkv: Make sense :)

protos/logging/logging.proto
irsdkv: @julianoes now Im looking at QGC implementation of [logging stream processing](https://github.com/mavlink/qgroundcontrol/blob/2f5324fd1745aa7ffa8ab1d477e2c8f1cad61c1c/src/Vehicle/MAVLinkLogManager.cc#L238) and I see that may exist at least two ways:

- Transieve in gRPC `first_message_offset`, `data` and `seguence` to full processing on the other side of gRPC.
- Create 'message LoggingMessage` and `message LoggingFileHeader' and implement some processing like QGC in MAVSDK.

I think that both ways have an advantages and disadvantages.

Which case seems like preferred for you? Or may be you have any other idea?

irsdkv: Or we can pasre `msg_id` in header and stream every messages as a custom gRPC `messages`

julianoes: > Create message LoggingFileHeader and message LoggingMessage (it will contain only message raw data) and implement some processing like QGC in MAVSDK

I think that makes sense.

julianoes: I would like to abstract it away as much as possible to have the easiest API.

julianoes:
Sorry, I didn't check this any more. How should we proceed @irsdkv? If you think this usable and useful like this, then we should probably get it in.

irsdkv:
Hm. I think need to implement some variant of logging in Backend core first.
And I think that it should be the some variant of  [QGC handling](https://github.com/mavlink/qgroundcontrol/blob/2f5324fd1745aa7ffa8ab1d477e2c8f1cad61c1c/src/Vehicle/MAVLinkLogManager.cc#L238) which does not compatible wit this Proto version. So this PR seems like outdated

julianoes:
@irsdkv I'm closing this for now because the logging plugin is not actually implemented.

PR #105, opened 2019-09-05
Initial MoCap service

irsdkv:
This is a draft to MoCap plugin API implementation.
I allowed myself to move slightly away from the usual API. May be you will find it useful. 
Please let me know is that interface ok.

Links:
[External Positioning](https://dev.px4.io/v1.9.0/en/ros/external_position_estimation.html)
[VISION_POSITION_ESTIMATE](https://mavlink.io/en/messages/common.html#VISION_POSITION_ESTIMATE)
[MAVSDK PR](https://github.com/mavlink/MAVSDK/pull/850)

### DIFF
--- a/protos/mocap/mocap.proto
+++ b/protos/mocap/mocap.proto
@@ -131,6 +131,7 @@ message MocapResult {
         SUCCESS = 1; // Request succeeded
         NO_SYSTEM = 2; // No system is connected
         CONNECTION_ERROR = 3; // Connection error
+        INVALID_REQUEST_DATA = 4; // Invalid request data
     }
 
     Result result = 1; // Result enum value
--- a/protos/mocap/mocap.proto
+++ b/protos/mocap/mocap.proto
@@ -33,7 +33,7 @@ message VisionPositionEstimate {
     float roll = 5; // Roll angle (rad)
     float pitch = 6; // Pitch angle (rad)
     float yaw = 7; // Yaw angle (rad)
-    repeated float covariance = 8; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle. Leave empty if unknown.
+    repeated float covariance = 8; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle. Set first to NaN if unknown.
     uint32 reset_counter = 9; // Estimate reset counter.
 }
 
@@ -52,7 +52,7 @@ message AttPosMocap {
     float x = 3; // X position (NED)
     float y = 4; // Y position (NED)
     float z = 5; // Z position (NED)
-    repeated float covariance = 6; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle.  Leave empty if unknown.
+    repeated float covariance = 6; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle. Set first to NaN if unknown.
 }
 
 
@@ -75,20 +75,12 @@ message Odometry {
       MAV_FRAME_GLOBAL_INT = 5; // Global (WGS84) coordinate frame (scaled) + MSL altitude.
       MAV_FRAME_GLOBAL_RELATIVE_ALT_INT = 6; // Global (WGS84) coordinate frame (scaled) + altitude relative to the home position.
       MAV_FRAME_LOCAL_OFFSET_NED = 7; // Offset to the current local frame. Anything expressed in this frame should be added to the current local frame position.
-      MAV_FRAME_BODY_NED = 8; // Setpoint in body NED frame. WARNING: DEPRECATED: Replaced by MAV_FRAME_BODY_FRD (2019-08).
-      MAV_FRAME_BODY_OFFSET_NED = 9; // Offset in body NED frame. WARNING: DEPRECATED: Replaced by MAV_FRAME_BODY_FRD (2019-08).
       MAV_FRAME_GLOBAL_TERRAIN_ALT = 10; // Global (WGS84) coordinate frame with AGL altitude (at the waypoint coordinate).
       MAV_FRAME_GLOBAL_TERRAIN_ALT_INT = 11; // Global (WGS84) coordinate frame (scaled) with AGL altitude (at the waypoint coordinate).
-      MAV_FRAME_BODY_FRD = 12; // Body fixed frame of reference, Z-down (x: forward, y: right, z: down).
+      MAV_FRAME_BODY_FRD = 12; // Body fixed frame of reference, Z-down (x: forward, y: right, z: down). Replacement for MAV_FRAME_BODY_NED, MAV_FRAME_BODY_OFFSET_NED.
       MAV_FRAME_BODY_FLU = 13; // Body fixed frame of reference, Z-up (x: forward, y: left, z: up).
-      MAV_FRAME_MOCAP_NED = 14; // Odometry local coordinate frame of data given by a motion capture system, Z-down (x: north, y: east, z: down). WARNING: DEPRECATED: Replaced by MAV_FRAME_LOCAL_FRD (2019-08).
-      MAV_FRAME_MOCAP_ENU = 15; // Odometry local coordinate frame of data given by a motion capture system, Z-up (x: east, y: north, z: up). WARNING: DEPRECATED: Replaced by MAV_FRAME_LOCAL_FLU (2019-08).
-      MAV_FRAME_VISION_NED = 16; // Odometry local coordinate frame of data given by a vision estimation system, Z-down (x: north, y: east, z: down). WARNING: DEPRECATED: Replaced by MAV_FRAME_LOCAL_FRD (2019-08).
-      MAV_FRAME_VISION_ENU = 17; // Odometry local coordinate frame of data given by a vision estimation system, Z-up (x: east, y: north, z: up). WARNING: DEPRECATED: Replaced by MAV_FRAME_LOCAL_FLU (2019-08).
-      MAV_FRAME_ESTIM_NED = 18; // Odometry local coordinate frame of data given by an estimator running onboard the vehicle, Z-down (x: north, y: east, z: down). WARNING: DEPRECATED: Replaced by MAV_FRAME_LOCAL_FRD (2019-08).
-      MAV_FRAME_ESTIM_ENU = 19; // Odometry local coordinate frame of data given by an estimator running onboard the vehicle, Z-up (x: east, y: noth, z: up). WARNING: DEPRECATED: Replaced by MAV_FRAME_LOCAL_FLU (2019-08).
-      MAV_FRAME_LOCAL_FRD = 20; // Forward, Right, Down coordinate frame. This is a local frame with Z-down and arbitrary F/R alignment (i.e. not aligned with NED/earth frame).
-      MAV_FRAME_LOCAL_FLU = 21; // Forward, Left, Up coordinate frame. This is a local frame with Z-up and arbitrary F/L alignment (i.e. not aligned with ENU/earth frame).
+      MAV_FRAME_LOCAL_FRD = 20; // Forward, Right, Down coordinate frame. This is a local frame with Z-down and arbitrary F/R alignment (i.e. not aligned with NED/earth frame). Replacement for MAV_FRAME_MOCAP_NED, MAV_FRAME_VISION_NED, MAV_FRAME_ESTIM_NED.
+      MAV_FRAME_LOCAL_FLU = 21; // Forward, Left, Up coordinate frame. This is a local frame with Z-up and arbitrary F/L alignment (i.e. not aligned with ENU/earth frame). Replacement for MAV_FRAME_MOCAP_ENU, MAV_FRAME_VISION_ENU, MAV_FRAME_ESTIM_ENU.
     }
 
     // Enumeration of estimator types
@@ -117,8 +109,8 @@ message Odometry {
     float rollspeed = 11; // Roll angular speed (rad/s).
     float pitchspeed = 12; // Pitch angular speed (rad/s).
     float yawspeed = 13; // Yaw angular speed (rad/s).
-    repeated float pose_covariance = 14; // Row-major representation of a 6x6 pose cross-covariance matrix upper right triangle. Leave empty if unknown.
-    repeated float velocity_covariance = 15; // Row-major representation of a 6x6 velocity cross-covariance matrix upper right triangle. Leave empty if unknown.
+    repeated float pose_covariance = 14; // Row-major representation of a 6x6 pose cross-covariance matrix upper right triangle. Set first to NaN if unknown.
+    repeated float velocity_covariance = 15; // Row-major representation of a 6x6 velocity cross-covariance matrix upper right triangle. Set first to NaN if unknown.
     uint32 reset_counter = 16; // Estimate reset counter.
     MavEstimatorType estimator_type = 17; // Type of estimator that is providing the odometry (not used at 2019-09).
 }
--- a/protos/mocap/mocap.proto
+++ b/protos/mocap/mocap.proto
@@ -1,0 +1,46 @@
+syntax = "proto3";
+
+package mavsdk.rpc.mocap;
+
+option java_package = "io.mavsdk.mocap";
+option java_outer_classname = "MocapProto";
+
+/**
+ * TODO
+ */
+service MocapService {
+    /*
+     * TODO
+     */
+    rpc CreateVisionPositionEstimateStream(stream VisionPositionEstimate) returns(stream VisionPositionEstimateResponse) {}
+}
+
+// Global position/attitude estimate from a vision source.
+message VisionPositionEstimate {
+    float x = 1; // Global X position (m)
+    float y = 2; // Global Y position (m)
+    float z = 3; // Global Z position (m)
+    float roll = 4; // Roll angle (rad)
+    float pitch = 5; // Pitch angle (rad)
+    float yaw = 6; // Yaw angle (rad)
+    repeated float covariance = 7; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle
+    uint32 reset_counter = 8; // Estimate reset counter.
+}
+
+message VisionPositionEstimateResponse {
+  MocapResult mocap_result = 1;
+}
+
+// Result type.
+message MocapResult {
+    // Possible results returned for mocap requests
+    enum Result {
+        UNKNOWN = 0; // Unknown error
+        SUCCESS = 1; // Request succeeded
+        NO_SYSTEM = 2; // No system is connected
+        CONNECTION_ERROR = 3; // Connection error
+    }
+
+    Result result = 1; // Result enum value
+    string result_str = 2; // Human-readable English string describing the result
+}
--- a/protos/mocap/mocap.proto
+++ b/protos/mocap/mocap.proto
@@ -13,6 +13,7 @@ service MocapService {
      * TODO
      */
     rpc CreateVisionPositionEstimateStream(stream VisionPositionEstimateRequest) returns(stream VisionPositionEstimateResponse) {}
+    rpc CreateAttPosMocapStream(stream AttPosMocapRequest) returns(stream AttPosMocapResponse) {}
 }
 
 message VisionPositionEstimateRequest {
@@ -24,14 +25,33 @@ message VisionPositionEstimateResponse {
 
 // Global position/attitude estimate from a vision source.
 message VisionPositionEstimate {
-    float x = 1; // Global X position (m)
-    float y = 2; // Global Y position (m)
-    float z = 3; // Global Z position (m)
-    float roll = 4; // Roll angle (rad)
-    float pitch = 5; // Pitch angle (rad)
-    float yaw = 6; // Yaw angle (rad)
-    repeated float covariance = 7; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle
-    uint32 reset_counter = 8; // Estimate reset counter.
+    uint64 time_usec = 1; // Position frame timestamp UNIX Epoch time (0 to use Backend timestamp)
+    float x = 2; // Global X position (m)
+    float y = 3; // Global Y position (m)
+    float z = 4; // Global Z position (m)
+    float roll = 5; // Roll angle (rad)
+    float pitch = 6; // Pitch angle (rad)
+    float yaw = 7; // Yaw angle (rad)
+    repeated float covariance = 8; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle
+    uint32 reset_counter = 9; // Estimate reset counter.
+}
+
+
+message AttPosMocapRequest {
+    AttPosMocap att_pos_mocap = 1;
+}
+message AttPosMocapResponse {
+  MocapResult mocap_result = 1;
+}
+
+// Motion capture attitude and position
+message AttPosMocap {
+    uint64 time_usec = 1; // Position frame timestamp UNIX Epoch time (0 to use Backend timestamp)
+    repeated float q = 2; // Attitude quaternion (w, x, y, z order, zero-rotation is 1, 0, 0, 0)
+    float x = 3; // X position (NED)
+    float y = 4; // Y position (NED)
+    float z = 5; // Z position (NED)
+    repeated float covariance = 6; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle
 }
 
 // Result type.
--- a/protos/mocap/mocap.proto
+++ b/protos/mocap/mocap.proto
@@ -12,7 +12,14 @@ service MocapService {
     /*
      * TODO
      */
-    rpc CreateVisionPositionEstimateStream(stream VisionPositionEstimate) returns(stream VisionPositionEstimateResponse) {}
+    rpc CreateVisionPositionEstimateStream(stream VisionPositionEstimateRequest) returns(stream VisionPositionEstimateResponse) {}
+}
+
+message VisionPositionEstimateRequest {
+    VisionPositionEstimate vision_position_estimate = 1;
+}
+message VisionPositionEstimateResponse {
+  MocapResult mocap_result = 1;
 }
 
 // Global position/attitude estimate from a vision source.
@@ -25,10 +32,6 @@ message VisionPositionEstimate {
     float yaw = 6; // Yaw angle (rad)
     repeated float covariance = 7; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle
     uint32 reset_counter = 8; // Estimate reset counter.
-}
-
-message VisionPositionEstimateResponse {
-  MocapResult mocap_result = 1;
 }
 
 // Result type.
--- a/protos/mocap/mocap.proto
+++ b/protos/mocap/mocap.proto
@@ -14,6 +14,7 @@ service MocapService {
      */
     rpc CreateVisionPositionEstimateStream(stream VisionPositionEstimateRequest) returns(stream VisionPositionEstimateResponse) {}
     rpc CreateAttPosMocapStream(stream AttPosMocapRequest) returns(stream AttPosMocapResponse) {}
+    rpc CreateOdometryStream(stream OdometryRequest) returns(stream OdometryResponse) {}
 }
 
 message VisionPositionEstimateRequest {
@@ -32,7 +33,7 @@ message VisionPositionEstimate {
     float roll = 5; // Roll angle (rad)
     float pitch = 6; // Pitch angle (rad)
     float yaw = 7; // Yaw angle (rad)
-    repeated float covariance = 8; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle
+    repeated float covariance = 8; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle. Leave empty if unknown.
     uint32 reset_counter = 9; // Estimate reset counter.
 }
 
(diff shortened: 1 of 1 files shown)


irsdkv:
I'm founded that MAVSDK-Python generator don't agree with my API :)

JonasVautherin:
> I'm founded that MAVSDK-Python protoc generator don't agree with my API :)

Is that because of the `repeated float`? I could imagine that I did not implemented everything related to the repeated types :-).

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,49 @@
...
+    float z = 3; // Global Z position (m)
+    float roll = 4; // Roll angle (rad)
+    float pitch = 5; // Pitch angle (rad)
+    float yaw = 6; // Yaw angle (rad)
+    repeated float covariance = 7; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle
+    uint32 reset_counter = 8; // Estimate reset counter.
+}
+
+// Result type.
+message MocapResult {
```
JonasVautherin: I don't know the Mocap interface. Does that have to be a stream? Sounds like it gives the state regarding "creating a position estimate stream from the client", doesn't it?

irsdkv:
> > I'm founded that MAVSDK-Python protoc generator don't agree with my API :)
> 
> Oh, because you send a stream. Yep, that's something new . But fortunately the last of 4 types of RPC supported by gRPC .

Yes, [API change](https://github.com/mavlink/MAVSDK-Proto/pull/105/commits/69c78d0d732198a55659b6c7287cecec72c3465b) resolved this problem. I was not tested yet, but generation works ok :)

protos/mocap/mocap.proto
irsdkv: > creating a position estimate stream from the client
Yes, it is
I think that stream is more suitable for these purposes. Position estimate frame should have at least 50Hz frequency, so I think that Stream exactly what is needed.

irsdkv: But on the other side if we want to know position set request result may be will be more suitable single request-response scheme. 
But may be that will be slowly than stream?

JonasVautherin: I'm not sure I get that. Is it normal to have a stream of `MocapResult`? When do I receive `SUCCESS`? When I open the connection, to tell me that I can now start sending values? Every time I send a new value? When I'm done sending values and the connection was closed successfully?

JonasVautherin: Ok I think I get it: we continuously send new position estimates, and that can fail if the connection somehow fails. But it's not like PX4 sends a `SUCCESS` status for each message. We just assume that if we could send the message, then it is a `SUCCESS`.

Until now, receiving `SUCCESS` from the server has always meant "end of stream". The only one having "intermediate results" is the [calibration plugin](https://github.com/mavlink/MAVSDK-Proto/blob/master/protos/calibration/calibration.proto#L54-L56), which has `IN_PROGRESS` and `INSTRUCTION` (that I believe should be merged into a common `NEXT` message). So that we would have:

- `SUCCESS` means that the process is complete
- `NEXT` is an intermediate message
- Everything else is a failure

And we could consider using `NEXT` here as well. But I'm tempted to think that in this case, we actually only want to act on error, i.e. the server-side stream would not send anything unless it is an error. We could have a result for that, e.g. `OK` or `NOOP`, that would be returned by the C++ implementation, but that would be ignored by gRPC...

Not sure if what I'm saying is clear :sweat_smile:. Does it make any sense at all?

irsdkv: > We just assume that if we could send the message, then it is a `SUCCESS`.

So we need to define "normal" execution without response.
In that case we should have something like 
`rpc VpeStream (stream VpeRequest) returns(VpeResponse)`
But in gRPC we cant cancel client stream on server-side (with streaming from client to server) **and send error code as response**. May be I wrong, but I [don't find](https://grpc.io/docs/guides/error/) way to close client stream on server-side and send response code (like `VpeResponse`).

I suggest this solution:

 - RPC definition:
`rpc VpeStream(stream VpeRequest)  returns(stream VpeResponse)`

 - If received `VpeRequest` sent to UAV without errors - do not respond.

 - If received `VpeRequest` cant be send to UAV (connection lost or any other problem) - send to client `VpeResponse` with error code

 - After first `VpeResponse` code was sent - reply corresponding error on every next `VpeRequest` until connection restore

 - If connection has been restored - send `VpeResponse::SUCCESS" once and after that continue without `VpeResponse` until next communication error.

irsdkv: >  I think that the connection will be closed as soon as the server sends the response (needs to be confirmed).

Hm. I will check that. If it really is so it is what we need, yes.

> Anyway if you lose the connection to the system, all the messages you send from MAVDSK will fail.

Need to confirm. 
I perform one small test on MAVSDK. Connect (over PixRacer WiFi), disconnect (turn WiFi off) and connect again (turn on WiFi). Seems that backend can regain connection:
```
[03:07:48|Debug] Sending msg 0 (system_impl.cpp:495)
[03:07:49|Debug] Sending msg 0 (system_impl.cpp:495)
[03:07:50|Debug] Sending msg 0 (system_impl.cpp:495)
[03:07:51|Info ] heartbeats timed out (system_impl.cpp:305) # Disconnect
[03:07:51|Debug] Lost 3546392297223829815 (mavsdk_impl.cpp:388)
[03:07:51|Info ] System timed out [UUID: 3546392297223829815] (connection_initiator.h:39)
[03:08:03|Debug] Discovered 2 component(s) (UUID: 3546392297223829815) (system_impl.cpp:558)
[03:08:03|Debug] Sending msg 76 (system_impl.cpp:495)
[03:08:03|Debug] Sending msg 76 (system_impl.cpp:495)
[03:08:03|Debug] Sending msg 0 (system_impl.cpp:495)
[03:08:03|Debug] MAVLink: critical: Data link lost (system_impl.cpp:300)
[03:08:03|Debug] MAVLink: info: Data link regained (system_impl.cpp:300)  # Connect
[03:08:04|Debug] Sending msg 0 (system_impl.cpp:495)
```

> Is it important to know which one failed? I understand it more as some kind of unreliable channel: you always want the server to know the latest message, but you don't care about the ones that were missed. Isn't that correct?

I think that you are right. There is no reason to know which exactly message was lost.

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,69 @@
...
+message VisionPositionEstimateRequest {
+    VisionPositionEstimate vision_position_estimate = 1;
+}
+message VisionPositionEstimateResponse {
+  MocapResult mocap_result = 1;
+}
+
+// Global position/attitude estimate from a vision source.
+message VisionPositionEstimate {
+    uint64 time_usec = 1; // Position frame timestamp UNIX Epoch time (0 to use Backend timestamp)
```
irsdkv: This can be sensible for FMU:
[time delay](https://dev.px4.io/v1.9.0/en/ros/external_position_estimation.html#tuning-EKF2_EV_DELAY)

protos/mocap/mocap.proto
JonasVautherin: > Hm. I will check that. If it really is so it is what we need, yes.

Did you find time to confirm that? Would be helpful :-)

protos/mocap/mocap.proto
```diff
@@ -32,7 +33,7 @@ message VisionPositionEstimate {
     float roll = 5; // Roll angle (rad)
     float pitch = 6; // Pitch angle (rad)
     float yaw = 7; // Yaw angle (rad)
-    repeated float covariance = 8; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle
+    repeated float covariance = 8; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle. Leave empty if unknown.
```
JonasVautherin: I think that it will [default to "0"](https://developers.google.com/protocol-buffers/docs/proto3#default) if we don't set it:

> For numeric types, the default value is zero.

I believe we should set it to `nan` if we don't want it. For instance [this example](https://github.com/mavlink/MAVSDK-Python/blob/master/examples/mission.py#L22) sets it to `nan`. A further step will be to allow for "optional" values, so that the user does not have to set it to `nan` explicitly, but we don't have that yet :blush:.

protos/mocap/mocap.proto
```diff
@@ -51,7 +52,75 @@ message AttPosMocap {
...
+      MAV_FRAME_VISION_ENU = 17; // Odometry local coordinate frame of data given by a vision estimation system, Z-up (x: east, y: north, z: up). WARNING: DEPRECATED: Replaced by MAV_FRAME_LOCAL_FLU (2019-08).
+      MAV_FRAME_ESTIM_NED = 18; // Odometry local coordinate frame of data given by an estimator running onboard the vehicle, Z-down (x: north, y: east, z: down). WARNING: DEPRECATED: Replaced by MAV_FRAME_LOCAL_FRD (2019-08).
+      MAV_FRAME_ESTIM_ENU = 19; // Odometry local coordinate frame of data given by an estimator running onboard the vehicle, Z-up (x: east, y: noth, z: up). WARNING: DEPRECATED: Replaced by MAV_FRAME_LOCAL_FLU (2019-08).
+      MAV_FRAME_LOCAL_FRD = 20; // Forward, Right, Down coordinate frame. This is a local frame with Z-down and arbitrary F/R alignment (i.e. not aligned with NED/earth frame).
+      MAV_FRAME_LOCAL_FLU = 21; // Forward, Left, Up coordinate frame. This is a local frame with Z-up and arbitrary F/L alignment (i.e. not aligned with ENU/earth frame).
```
julianoes: Can we try not to expose all the MAVLink baggage (aka all the unused frames) and only expose a supported and very clear subset?

irsdkv: Hm.
This is good idea. But I'm afraid that some systems or users can use or maintain old revisions of PX4. 
Is it not critical?

protos/mocap/mocap.proto
irsdkv: Why not :)
I will implement that.

protos/mocap/mocap.proto
julianoes: We only need to support the current release and master usually. For the future we might want to be somewhat backwards compatible but for now I would not worry about that, especially since it's a new feature in MAVSDK anyway.

protos/mocap/mocap.proto
irsdkv: Seems like we cant close stream with Server side and send Response.

We only can use 

`response->set_allocated_mocap_result(response_val);`

This result will be sent when client call [`writer->WritesDone();` ](https://grpc.io/docs/tutorials/basic/cpp/).
So at this moment I think that most easy and clear way to implement that is both-side-streaming.

protos/mocap/mocap.proto
irsdkv: Done: https://github.com/mavlink/MAVSDK-Proto/pull/105/commits/624e16b4e2213a3fd244c350236c20ed8dc45e94#diff-32ec9342f13610cc1fb904668f21d7ccR36
Backend impl: https://github.com/mavlink/MAVSDK/pull/850/commits/d1b917c27e2807f94a92f3a819c097573c204f33#diff-8d123a99eeadace7e5baafa7fa5d6105R206

protos/mocap/mocap.proto
irsdkv: Done: https://github.com/mavlink/MAVSDK-Proto/pull/105/commits/624e16b4e2213a3fd244c350236c20ed8dc45e94

protos/mocap/mocap.proto
JonasVautherin: > Seems like we cant close stream on Server side with sending Response.

I don't understand what that means. When the server sends `return grpc::Status::OK;` in your C++ code, it most likely closes the client-side stream. Right?

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,131 @@
...
+message VisionPositionEstimate {
+    uint64 time_usec = 1; // Position frame timestamp UNIX Epoch time (0 to use Backend timestamp)
+    float x = 2; // Global X position (m)
+    float y = 3; // Global Y position (m)
+    float z = 4; // Global Z position (m)
+    float roll = 5; // Roll angle (rad)
+    float pitch = 6; // Pitch angle (rad)
+    float yaw = 7; // Yaw angle (rad)
+    repeated float covariance = 8; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle. Set first to NaN if unknown.
+    uint32 reset_counter = 9; // Estimate reset counter.
```
julianoes: How does the reset counter work? Is it something that could happen MAVSDK internally or is this something that the API user needs?

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,131 @@
+syntax = "proto3";
+
+package mavsdk.rpc.mocap;
+
+option java_package = "io.mavsdk.mocap";
+option java_outer_classname = "MocapProto";
+
+/**
+ * TODO
```
julianoes: Please describe at least that Mocap means Motion Capture :smile:.

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,131 @@
...
+
+message AttPosMocapRequest {
+    AttPosMocap att_pos_mocap = 1;
+}
+message AttPosMocapResponse {
+  MocapResult mocap_result = 1;
+}
+
+// Motion capture attitude and position
+message AttPosMocap {
```
julianoes: I would call it AttitudePositionMocap. In general I vote for more verbose but clearer naming.

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,131 @@
...
+    float z = 5; // Z position (NED)
+    repeated float covariance = 6; // Row-major representation of pose 6x6 cross-covariance matrix upper right triangle. Set first to NaN if unknown.
+}
+
+
+message OdometryRequest {
+    Odometry odometry = 1;
+}
+message OdometryResponse {
+  MocapResult mocap_result = 1;
```
julianoes: Indenting is not consistent in this file.

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,131 @@
...
+      MAV_FRAME_GLOBAL_TERRAIN_ALT = 10; // Global (WGS84) coordinate frame with AGL altitude (at the waypoint coordinate).
+      MAV_FRAME_GLOBAL_TERRAIN_ALT_INT = 11; // Global (WGS84) coordinate frame (scaled) with AGL altitude (at the waypoint coordinate).
+      MAV_FRAME_BODY_FRD = 12; // Body fixed frame of reference, Z-down (x: forward, y: right, z: down). Replacement for MAV_FRAME_BODY_NED, MAV_FRAME_BODY_OFFSET_NED.
+      MAV_FRAME_BODY_FLU = 13; // Body fixed frame of reference, Z-up (x: forward, y: left, z: up).
+      MAV_FRAME_LOCAL_FRD = 20; // Forward, Right, Down coordinate frame. This is a local frame with Z-down and arbitrary F/R alignment (i.e. not aligned with NED/earth frame). Replacement for MAV_FRAME_MOCAP_NED, MAV_FRAME_VISION_NED, MAV_FRAME_ESTIM_NED.
+      MAV_FRAME_LOCAL_FLU = 21; // Forward, Left, Up coordinate frame. This is a local frame with Z-up and arbitrary F/L alignment (i.e. not aligned with ENU/earth frame). Replacement for MAV_FRAME_MOCAP_ENU, MAV_FRAME_VISION_ENU, MAV_FRAME_ESTIM_ENU.
```
julianoes: Which frames do we actually need? Can we reduce it to e.g. 3? And which ones are actually supported by PX4?

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,131 @@
...
+    float y = 5; // Y position.
+    float z = 6; // Z position.
+    repeated float q = 7; // Quaternion components, w, x, y, z (1 0 0 0 is the null-rotation).
+    float vx = 8; // X linear speed (m/s).
+    float vy = 9; // Y linear speed (m/s).
+    float vz = 10; // Z linear speed (m/s).
+    float rollspeed = 11; // Roll angular speed (rad/s).
+    float pitchspeed = 12; // Pitch angular speed (rad/s).
+    float yawspeed = 13; // Yaw angular speed (rad/s).
+    repeated float pose_covariance = 14; // Row-major representation of a 6x6 pose cross-covariance matrix upper right triangle. Set first to NaN if unknown.
```
julianoes: Are there protobuf types that we could re-use here?

irsdkv:
Hm.
I found that Python generator doesn't work fine outside naming like `SetWhatever(SetWhateverRequest) returns([stream] SetWhateverResponse)`.
It is not ok for testing and using.
At this time I tested single request-response scheme and it work ok in around of 30Hz on 433MHz radio.
Now I think that fastest and clearest way to implement this module will be using this "old" way (single request-response).
What do you think about it?

protos/mocap/mocap.proto
irsdkv: I founded thar these types not used in PX4 FW currently:
- MAV_FRAME_LOCAL_ENU
- MAV_FRAME_LOCAL_OFFSET_NED (PX4/Firmware#5407 related)
- MAV_FRAME_GLOBAL_TERRAIN_ALT
- MAV_FRAME_GLOBAL_TERRAIN_ALT_INT
- MAV_FRAME_LOCAL_FRD
- MAV_FRAME_LOCAL_FLU

I think most of it we can exclude from proto (except `MAV_FRAME_LOCAL_OFFSET_NED`, may be)

protos/mocap/mocap.proto
irsdkv: We can use `mavsdk.rpc.telemetry.Quaternion` and `mavsdk.rpc.telemetry.AngularVelocityBody` for example. But I think that in this case we need to place out reused types in one common proto file. Because of it will little bit confusing if user should import, for example `telemetry` to use `mocap`. And also in protos Quaternion used in Camera and Telemetry both now. 
May be need to do some replacing common types in future...
At this time I will try to place some data types in `mocap.proto`. May be it will be ok for now.

julianoes: Ok, I think for now we duplicate the types as you have it here. It's easier for now.

protos/mocap/mocap.proto
irsdkv: `reset_counter` deleted

protos/mocap/mocap.proto
irsdkv: `Quaternion`, `Position` and `Covariance` reused now

protos/mocap/mocap.proto
irsdkv: At this time I think that both-side streaming too complicated because of python-generator cant generate appropriate code. 
Returned to single Request-Response scheme (https://github.com/mavlink/MAVSDK-Proto/pull/105#issuecomment-534513268).
It is a pity that such a long discussion could not be used so far :(
I will check that if will necessary to come back to client-server streaming in this module

irsdkv:
Small test result:
This implementation (with [this commit](https://github.com/mavlink/MAVSDK/pull/850/commits/805c449f044b6facdd959cb878ed57601bbe2385)) has reached more than 250Hz frequency of sending VisualPositionEstimate using MAVSDK-Python and ESP8266-12e link.

julianoes:
Oh nice!

JonasVautherin:
> I found that Python generator doesn't work fine outside naming like SetWhatever(SetWhateverRequest) returns([stream] SetWhateverResponse).

Yes, because we would need to extend the generator to support client streams.

@julianoes: do you think we should go for this implementation right now, and possibly move to a client-side stream later if that gets contributed?

@irsdkv: how would you represent a client stream in python? A server stream becomes a generator, but I'm not sure what a client stream should be :blush:

irsdkv:
> @julianoes: do you think we should go for this implementation right now, and possibly move to a client-side stream later if that gets contributed?

One day more for me please :)
Now I'm try to implement Telemetry Odometry stream. I would try to reach more compatibility for `Odometry` types in `Mocap` and `Telemetry` (may be in future they will go outside plugins in common types header).
 
> @irsdkv: how would you represent a client stream in python? A server stream becomes a generator, but I'm not sure what a client stream should be

Yes, client-side stream is not generated now (absolutely non-functional generated code, if I understand it right). I tried to implement it, but it complicated for me now.

JonasVautherin:
> So I have no idea how client stream should be implemented :)

I did not mean how the protoc plugin should be implemented, but the actual python code. Say `SetVisionPositionEstimate` is a client stream, that could be something like:

```python
position_estimate_subject = drone.mocap.set_vision_position_estimate()

while True:
    position_estimate_subject.post("the new value I want to send to the server")

position_estimate_subject.close()
```

For instance. Well, adapted for asyncio, somehow. What do you think?

irsdkv:
> For instance. Well, adapted for asyncio, somehow. What do you think?

Hm. May be I will try it later.

irsdkv:
> ```python
> position_estimate_subject = drone.mocap.set_vision_position_estimate()
> 
> while True:
>     position_estimate_subject.post("the new value I want to send to the server")
> 
> position_estimate_subject.close()
> ```

I definitely try it later (may be next week). But may be make sense go for this implementation at this moment?
So we will have some worked implementation as the reference.

irsdkv:
So, tests showed that the current implementation is working correctly.

OTA `VisualPositionEstimate` streaming:
image

"Fake" position streaming:
image

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,153 @@
...
+}
+message SetOdometryResponse {
+    MocapResult mocap_result = 1;
+}
+
+// Global position/attitude estimate from a vision source.
+message VisionPositionEstimate {
+    uint64 time_usec = 1; // Position frame timestamp UNIX Epoch time (0 to use Backend timestamp)
+    Position position = 2; // Global position (m)
+    float roll = 3; // Roll angle (rad)
```
julianoes: Please add the unit.

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,153 @@
...
+    Quaternion q = 2; // Attitude quaternion (w, x, y, z order, zero-rotation is 1, 0, 0, 0)
+    Position position = 3; // Position (NED)
+    Covariance covariance = 6; // Pose cross-covariance matrix.
+}
+
+// Odometry message to communicate odometry information with an external interface.
+message Odometry {
+    // Mavlink frame id
+    enum MavFrame {
+        MAV_FRAME_GLOBAL = 0; // Global (WGS84) coordinate frame + MSL altitude.
```
julianoes: Please only expose frames that are known to be supported and working for PX4.

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,153 @@
...
+        MAV_ESTIMATOR_TYPE_LIDAR = 7; // Estimator based on lidar sensor input.
+        MAV_ESTIMATOR_TYPE_AUTOPILOT = 8; // Estimator on autopilot.
+    }
+
+    uint64 time_usec = 1; // Timestamp (0 to use Backend timestamp).
+    MavFrame frame_id = 2; // Coordinate frame of reference for the pose data.
+    MavFrame child_frame_id = 3;  // Coordinate frame of reference for the velocity in free space (twist) data.
+    Position position = 4; // Position.
+    Quaternion q = 5; // Quaternion components, w, x, y, z (1 0 0 0 is the null-rotation).
+    float vx = 6; // X linear speed (m/s).
```
julianoes: Units.

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,153 @@
...
+    float vy = 7; // Y linear speed (m/s).
+    float vz = 8; // Z linear speed (m/s).
+    float rollspeed = 9; // Roll angular speed (rad/s).
```
julianoes: Units.

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,153 @@
...
+    float pitchspeed = 10; // Pitch angular speed (rad/s).
+    float yawspeed = 11; // Yaw angular speed (rad/s).
+    Covariance pose_covariance = 12; // Pose cross-covariance matrix.
+    Covariance velocity_covariance = 13; // Velocity cross-covariance matrix.
+    MavEstimatorType estimator_type = 14; // Type of estimator that is providing the odometry (not used at 2019-09).
+}
+
+// Position type
+message Position {
+    float x = 1; // X position.
```
julianoes: Unit.

protos/mocap/mocap.proto
JonasVautherin: (e.g. `roll_rad` similar to [here](https://github.com/mavlink/MAVSDK-Proto/blob/master/protos/telemetry/telemetry.proto#L196))

JonasVautherin:
> I definitely try it later (may be next week). But may be make sense go for this implementation at this moment?

Sure, let's do that.

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,162 @@
...
+        UNKNOWN = 0; // Unknown type of the estimator.
+        NAIVE = 1; // This is a naive estimator without any real covariance feedback.
+        TYPE_VISION = 2; // Computer vision based estimate. Might be up to scale.
+        VIO = 3; // Visual-inertial estimate.
+        GPS = 4; // Plain GPS estimate.
+        INS = 5; // Estimator integrating GPS and inertial sensing.
+        MOCAP = 6; // Estimate from external motion capturing system.
+        LIDAR = 7; // Estimator based on lidar sensor input.
+        AUTOPILOT = 8; // Estimator on autopilot.
+    }
```
irsdkv: At this time field `estimator_type` does not used in PX4.
But I think that ability to use it in default plugin will be helpful for MoCap and VIO development.

irsdkv: At this time field `estimator_type` does not used in PX4.
But I think that ability to use it in default plugin will be helpful for MoCap and VIO development.

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,164 @@
...
+message Odometry {
+    // Mavlink frame id
+    enum MavFrame {
+        STUB = 0; // Stub for Protobuf
+        BODY_FRD = 12; // Body fixed frame of reference, Z-down (x: forward, y: right, z: down). Replacement for MAV_FRAME_BODY_NED, MAV_FRAME_BODY_OFFSET_NED.
+        BODY_FLU = 13; // Body fixed frame of reference, Z-up (x: forward, y: left, z: up).
+        MOCAP_NED = 14; // Odometry local coordinate frame of data given by a motion capture system, Z-down (x: north, y: east, z: down).
+        VISION_NED = 16; // Odometry local coordinate frame of data given by a motion capture system, Z-up (x: east, y: north, z: up).
+        LOCAL_FRD = 20; // Forward, Right, Down coordinate frame. This is a local frame with Z-down and arbitrary F/R alignment (i.e. not aligned with NED/earth frame). Replacement for MAV_FRAME_MOCAP_NED, MAV_FRAME_VISION_NED, MAV_FRAME_ESTIM_NED.
+        LOCAL_FLU = 21; // Forward, Left, Up coordinate frame. This is a local frame with Z-up and arbitrary F/L alignment (i.e. not aligned with ENU/earth frame). Replacement for MAV_FRAME_MOCAP_ENU, MAV_FRAME_VISION_ENU, MAV_FRAME_ESTIM_ENU.
```
irsdkv: Will be in use
https://mavlink.io/en/messages/common.html#MAV_FRAME_LOCAL_FLU

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,164 @@
...
+// Odometry message to communicate odometry information with an external interface.
+message Odometry {
+    // Mavlink frame id
+    enum MavFrame {
+        STUB = 0; // Stub for Protobuf
+        BODY_FRD = 12; // Body fixed frame of reference, Z-down (x: forward, y: right, z: down). Replacement for MAV_FRAME_BODY_NED, MAV_FRAME_BODY_OFFSET_NED.
+        BODY_FLU = 13; // Body fixed frame of reference, Z-up (x: forward, y: left, z: up).
+        MOCAP_NED = 14; // Odometry local coordinate frame of data given by a motion capture system, Z-down (x: north, y: east, z: down).
+        VISION_NED = 16; // Odometry local coordinate frame of data given by a motion capture system, Z-up (x: east, y: north, z: up).
+        LOCAL_FRD = 20; // Forward, Right, Down coordinate frame. This is a local frame with Z-down and arbitrary F/R alignment (i.e. not aligned with NED/earth frame). Replacement for MAV_FRAME_MOCAP_NED, MAV_FRAME_VISION_NED, MAV_FRAME_ESTIM_NED.
```
irsdkv: Will be in use
https://mavlink.io/en/messages/common.html#MAV_FRAME_LOCAL_FRD

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,164 @@
...
+
+// Odometry message to communicate odometry information with an external interface.
+message Odometry {
+    // Mavlink frame id
+    enum MavFrame {
+        STUB = 0; // Stub for Protobuf
+        BODY_FRD = 12; // Body fixed frame of reference, Z-down (x: forward, y: right, z: down). Replacement for MAV_FRAME_BODY_NED, MAV_FRAME_BODY_OFFSET_NED.
+        BODY_FLU = 13; // Body fixed frame of reference, Z-up (x: forward, y: left, z: up).
+        MOCAP_NED = 14; // Odometry local coordinate frame of data given by a motion capture system, Z-down (x: north, y: east, z: down).
+        VISION_NED = 16; // Odometry local coordinate frame of data given by a motion capture system, Z-up (x: east, y: north, z: up).
```
irsdkv: May be need to use
https://github.com/PX4/Firmware/blob/07d656e971a72d1202651dfd3b4642736fb078d7/src/modules/simulator/simulator_mavlink.cpp#L1083

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// Odometry message to communicate odometry information with an external interface.
+message Odometry {
+    // Mavlink frame id
+    enum MavFrame {
+        STUB = 0; // Stub for Protobuf
+        BODY_FRD = 12; // Body fixed frame of reference, Z-down (x: forward, y: right, z: down). Replacement for MAV_FRAME_BODY_NED, MAV_FRAME_BODY_OFFSET_NED.
+        BODY_FLU = 13; // Body fixed frame of reference, Z-up (x: forward, y: left, z: up).
+        MOCAP_NED = 14; // Odometry local coordinate frame of data given by a motion capture system, Z-down (x: north, y: east, z: down).
```
irsdkv: https://github.com/PX4/Firmware/blob/c8e59c4e39c9f46633c07ea2d16d9d03706caf69/src/modules/mavlink/mavlink_receiver.cpp#L1279

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,164 @@
...
+    Covariance pose_covariance = 4; // Pose cross-covariance matrix.
+}
+
+// Odometry message to communicate odometry information with an external interface.
+message Odometry {
+    // Mavlink frame id
+    enum MavFrame {
+        STUB = 0; // Stub for Protobuf
+        BODY_FRD = 12; // Body fixed frame of reference, Z-down (x: forward, y: right, z: down). Replacement for MAV_FRAME_BODY_NED, MAV_FRAME_BODY_OFFSET_NED.
+        BODY_FLU = 13; // Body fixed frame of reference, Z-up (x: forward, y: left, z: up).
```
irsdkv: May be need to use
https://github.com/PX4/Firmware/pull/12293

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,164 @@
...
+    PositionBody position_body = 3; // Body Position (NED)
+    Covariance pose_covariance = 4; // Pose cross-covariance matrix.
+}
+
+// Odometry message to communicate odometry information with an external interface.
+message Odometry {
+    // Mavlink frame id
+    enum MavFrame {
+        STUB = 0; // Stub for Protobuf
+        BODY_FRD = 12; // Body fixed frame of reference, Z-down (x: forward, y: right, z: down). Replacement for MAV_FRAME_BODY_NED, MAV_FRAME_BODY_OFFSET_NED.
```
irsdkv: https://github.com/PX4/Firmware/blob/c8e59c4e39c9f46633c07ea2d16d9d03706caf69/src/modules/mavlink/mavlink_receiver.cpp#L1251

protos/mocap/mocap.proto
JonasVautherin: I'm not sure about that. If it is not used, it means that we expose something that will fail. I would say that we only publish things that work with an actual autopilot. For development, one can always use a branch, and merge that into master when it is working.

In other words, I would be in favor of having in our proto files only things that we can test against an actual autopilot. We can always extend later.

What do you think?

protos/mocap/mocap.proto
```diff
@@ -0,0 +1,164 @@
...
+    Quaternion q = 2; // Attitude quaternion (w, x, y, z order, zero-rotation is 1, 0, 0, 0)
+    PositionBody position_body = 3; // Body Position (NED)
+    Covariance pose_covariance = 4; // Pose cross-covariance matrix.
+}
+
+// Odometry message to communicate odometry information with an external interface.
+message Odometry {
+    // Mavlink frame id
+    enum MavFrame {
+        STUB = 0; // Stub for Protobuf
```
JonasVautherin: Why do we need a `STUB` for protobuf?

protos/mocap/mocap.proto
julianoes: I very much agree. Otherwise we end up with the same problems as MAVLink where you don't know what to set for many params and you don't know if they are even used, supported, or what effects they have.

protos/mocap/mocap.proto
julianoes: My question is which frames are (currently) supported by PX4. And then I would only support and expose those. If we need more frames, then support in PX4 would have to be added first, in my opinion.

protos/mocap/mocap.proto
irsdkv:
```bash
$ ./generate_from_protos.sh 
-------------------------------
 Generating pb and grpc.pb files
    * protoc --version: libprotoc 3.9.1
-------------------------------

mocap/mocap.proto:64:20: The first enum value must be zero in proto3.
```

protos/mocap/mocap.proto
irsdkv: Good clarification. I think you are right.

protos/mocap/mocap.proto
irsdkv: UPD
Already in use
https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_receiver.cpp#L1276

protos/mocap/mocap.proto
```diff
@@ -61,37 +61,18 @@ message Odometry {
...
-        GPS = 4; // Plain GPS estimate.
-        INS = 5; // Estimator integrating GPS and inertial sensing.
-        MOCAP = 6; // Estimate from external motion capturing system.
-        LIDAR = 7; // Estimator based on lidar sensor input.
-        AUTOPILOT = 8; // Estimator on autopilot.
     }
 
     uint64 time_usec = 1; // Timestamp (0 to use Backend timestamp).
     MavFrame frame_id = 2; // Coordinate frame of reference for the pose data.
-    MavFrame child_frame_id = 3;  // Coordinate frame of reference for the velocity in free space (twist) data.
```
irsdkv: Child frame removed because of currently supported only [`MAV_FRAME_BODY_FRD` ](https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_receiver.cpp#L1251)

protos/mocap/mocap.proto
JonasVautherin: Doesn't that mean that `BODY_FRD` should be `= 0` instead of `= 12`? See here: https://developers.google.com/protocol-buffers/docs/proto3#enum.

And then the next ones like `BODY_FLU` could be `1`, `2`, etc. Or does it conflict with something else?

irsdkv: I think that setting `BODY_FRD = 0` may be confusing for gRPC-client users.
For example in MAVSDK-Python:
I [manually changed generated code](https://github.com/mavlink/MAVSDK-Python/pull/111/files#diff-755143e2465cb1cce8758a454a7ccdd4R273-R274) to delete `STUB` and set real values of `enum` members.
This will not conflict for any user-defined code which operated with MAVLink-defined constants  (for example `MOCAP_NED` eq `14`).
And also, on the other hand, bindings for gRPC plugins to other languages will be produced with users which familiar with MAVSDK and MAVLink and they will can also remove stubs from final API.
What

JonasVautherin: How is `BODY_FRD = 0` confusing? I don't get that :confused:.

JonasVautherin: Because having to change all the templates for all the languages just to say "ignore any enum value called `STUB` with a value of "0" because that's a workaround we did" is quite a strong and non-intuitive thing. I'd like to be sure we need that before :sweat_smile:.

irsdkv: I mean that if some constant values like `BODY_FRD` will be differ from its values in MAVLink then it can seem like collision.
For example if in MAVSDK-Proto `BODY_FRD = 0` so in MAVSDK-Python (and other bindings) this constant also eq `0` as default, but in MAVLink `BODY_FRD` = `12`. 
User code (in python for example) may check if current frame is `Mocap.BODY_FRD` and expect `12`, but this constant eq `0`

irsdkv: > Because having to change all the templates for all the languages just to say "ignore any enum value called `STUB` with a value of "0" because that's a workaround we did" is quite a strong and non-intuitive thing. I'd like to be sure we need that before .


Yes, I think that this stub (:smile:) not a best solution.
But other way (dont use `STUB`) looks more potentially dangerous for me. 
I do not insist on this decision, but I think that in case of dont use `STUB` maintainers should manually change values in bindings like MAVSDK-Python.
And also `static_cast` will not work for casting ever more in `plugin_service_impl.h` :)

JonasVautherin: I don't think I agree on the fact that this is a reason for adding a very weird condition in all the generated code :sweat_smile:. This is adding confusion in the templates in all languages, and they are already not super easy to read. And on the other side, I don't see it as a good practice to rely on the id of an enum value.

For instance, I don't think that protobuf says in the specs that `BODY_FRD = 0` will always translate that ID in the resulting generated code. Maybe it works now in Python, but that may well be out of luck. Or is that somewhere in the specs?

JonasVautherin: > And also static_cast will not work for casting ever more in plugin_service_impl.h :)

Yes, `static_cast` is convenient, but it requires a unit test to make sure that it works. If it doesn't, then we need to write more verbose code, which I believe can still be auto-generated.

irsdkv: Few more info from [docs](https://developers.google.com/protocol-buffers/docs/proto3#default):

> For enums, the default value is the first defined enum value, which must be 0.

So `0` is default. I don't know in which cases uses default values. May be this also need to take into consideration.

JonasVautherin: That's the default value in case you don't specify it. In many cases we have `UNKNOWN = 0`. But also, we always specify a value in the code we generate. So I don't think that has any effect for us.

JonasVautherin: I would vote for this: we remove `STUB`, and we start the others at 0 (i.e. we don't care about the MAVLink enum ids). If then we see that this makes it very confusing for python users, we can change the python templates to make it more intuitive.

This way we don't propagate a weird `STUB` to all the other languages.

What do you think? :blush:

irsdkv: I think that this make sense. I will do corresponding changes.

PR #106, opened 2019-09-14
add landed state in telemetry

coolkesword:
Hi, I have added landed state support in telemetry, kindly note that the enum is with C++ scope, so I have to add a prefix LANDED_ to existing landed_state enum in order to differentiate with the UNKNOWN in FlightMode enum. 

I will try to add further support in MAVSDK and MAVSDK_python as well.

Thanks
coolkesword

### DIFF
--- a/protos/telemetry/telemetry.proto
+++ b/protos/telemetry/telemetry.proto
@@ -280,9 +280,9 @@ message ActuatorOutputStatus {
 
 // Landed State enumeration
 enum LandedState{
-    LANDED_UNKNOWN = 0;
-    LANDED_ON_GROUND = 1;
-    LANDED_IN_AIR = 2;
-    LANDED_TAKING_OFF = 3;
-    LANDED_LANDING = 4;
+    LANDED_STATE_UNKNOWN = 0;
+    LANDED_STATE_ON_GROUND = 1;
+    LANDED_STATE_IN_AIR = 2;
+    LANDED_STATE_TAKING_OFF = 3;
+    LANDED_STATE_LANDING = 4;
 }
--- a/protos/telemetry/telemetry.proto
+++ b/protos/telemetry/telemetry.proto
@@ -16,6 +16,8 @@ service TelemetryService {
     rpc SubscribeHome(SubscribeHomeRequest) returns(stream HomeResponse) {}
     // Subscribe to in-air updates.
     rpc SubscribeInAir(SubscribeInAirRequest) returns(stream InAirResponse) {}
+    // Subscribe to landed state updates
+    rpc SubscribeLandedState(SubscribeLandedStateRequest) returns(stream LandedStateResponse) {}
     // Subscribe to armed updates.
     rpc SubscribeArmed(SubscribeArmedRequest) returns(stream ArmedResponse) {}
     // Subscribe to 'attitude' updates (quaternion).
@@ -59,6 +61,11 @@ message HomeResponse {
 message SubscribeInAirRequest {}
 message InAirResponse {
     bool is_in_air = 1; // The next 'in-air' state
+}
+
+message SubscribeLandedStateRequest {}
+message LandedStateResponse {
+    LandedState landed_state = 1;
 }
 
 message SubscribeArmedRequest {}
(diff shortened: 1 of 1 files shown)


JonasVautherin:
Is that related to a discussion we had a while ago? It reminds me of something, but I can't seem to find the related issue :confused:.

Also, I don't get why we need the `LANDED_` prefix... can you elaborate on that?

coolkesword:
Hi, Jonas Vautherin, this is not related. 

The problem is that as follows when we don't add a prefix, you may also suggest a better way to overcome it:
telemetry.proto:299:5: "UNKNOWN" is already defined in "mavsdk.rpc.telemetry".
telemetry.proto:299:5: Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  Therefore, "UNKNOWN" must be unique within "mavsdk.rpc.telemetry", not just within "LandedState".
 -> [+] Generated protobuf and gRPC bindings for telemetry\

Thanks

JonasVautherin:
Right. And what information does that bring as compared to a combination of `InAir` and `FlightMode`?

coolkesword:
> Right. And what information does that bring as compared to a combination of `InAir` and `FlightMode`?

Do you mean that we can add the in air landed state to the flightmode enumeration? I think the method can work but a confusion will be made between the flight mode and landed state, and the two enumerations are defined separately in MAVSDK.

JonasVautherin:
I mean that you want those values: `ON_GROUND`, `IN_AIR`, `TAKING_OFF` , `LANDING`, and I believe I can already get them with the current implementation.

* If `InAir` is false, it means `ON_GROUND`. Otherwise, if:
    * FlightMode is `TAKEOFF`, it means that you are `TAKING_OFF`
    * FlightMode is `LAND`, it means that you are `LANDING`
    * `InAir` is true, it means `IN_AIR`

What is it that you cannot solve with `InAir` and `FlightMode`, then? In other words, why do you need to add a "landed state" in MAVSDK?

coolkesword:
Thanks for your clarification, I have got your meaning. The main reason that I have to create a landed state enum is because I found the C++ MAVSDK api supports landed state enquiry, as shown in:
https://mavsdk.mavlink.io/develop/en/api_reference/classmavsdk_1_1_telemetry.html

Hence, I thought it might be more user friendly if we have landed state API in python, although this is not a technical problem following your illustration.

Regards
coolkesword

> I mean that you want those values: `ON_GROUND`, `IN_AIR`, `TAKING_OFF` , `LANDING`, and I believe I can already get them with the current implementation.
> 
>     * If `InAir` is false, it means `ON_GROUND`. Otherwise, if:
>       
>       * FlightMode is `TAKEOFF`, it means that you are `TAKING_OFF`
>       * FlightMode is `LAND`, it means that you are `LANDING`
>       * `InAir` is true, it means `IN_AIR`
> 
> 
> What is it that you cannot solve with `InAir` and `FlightMode`, then? In other words, why do you need to add a "landed state" in MAVSDK?

JonasVautherin:
I am personally not in favor of having multiple ways to get the same data. But maybe that's not the case, and `LandedState` does bring valuable information.

@julianoes what's your take on that?

julianoes:
@JonasVautherin I generally agree but this was already added to the C++ side of things, so the discussion here is probably too late :thinking:.

JonasVautherin:
> so the discussion here is probably too late :thinking: 

Well not sure: it's not yet propagated to all the language bindings, and we could deprecate the C++ one. Having two implementations to bring the same result is prone to error, IMO. And it's adding noise in the code.

julianoes:
Ok, in that case we should switch to the new more detailed info, however, the old one is implemented everywhere so it will cause quite a bit of churn.

JonasVautherin:
You mean that "InAir" should become "LandedState"? In that case we can deprecate "InAir" and add "LandedState", and that's it. Right?

julianoes:
Exactly.

JonasVautherin:
Ok! @coolkesword, the PR is back in the game :blush:. Sorry for the delay, I needed to clarify that!

protos/telemetry/telemetry.proto
```diff
@@ -270,3 +277,12 @@ message ActuatorOutputStatus {
     uint32 active = 1;
     repeated float actuator = 2;
 }
+
+// Landed State enumeration
+enum LandedState{
+    LANDED_UNKNOWN = 0;
+    LANDED_ON_GROUND = 1;
+    LANDED_IN_AIR = 2;
```
JonasVautherin: `LANDED_IN_AIR` sounds weird to me. Could we instead use `LANDED_STATE` as a prefix? `LANDED_STATE_UNKNOWN`, `LANDED_STATE_IN_AIR`, ...

coolkesword: Sure, LANDED_STATE prefix is perfectly fine. I have also made some changes to MAVSDK and mavsdk python to enable the landed_state enum, hopefully I can contribute accordingly after the PR is accepted. Thanks

PR #107, opened 2019-09-26
Telemetry Odometry stream added

irsdkv:
Due to https://github.com/mavlink/MAVSDK/pull/866

### DIFF
--- a/protos/telemetry/telemetry.proto
+++ b/protos/telemetry/telemetry.proto
@@ -44,8 +44,11 @@ service TelemetryService {
     rpc SubscribeRcStatus(SubscribeRcStatusRequest) returns(stream RcStatusResponse) {}
     // Subscribe to 'status text' updates.
     rpc SubscribeStatusText(SubscribeStatusTextRequest) returns(stream StatusTextResponse) {}
+    // Subscribe to 'actuator control target' updates.
     rpc SubscribeActuatorControlTarget(SubscribeActuatorControlTargetRequest) returns(stream ActuatorControlTargetResponse) {}
+    // Subscribe to 'actuator output status' updates.
     rpc SubscribeActuatorOutputStatus(SubscribeActuatorOutputStatusRequest) returns(stream ActuatorOutputStatusResponse) {}
+    // Subscribe to 'odometry' updates.
     rpc SubscribeOdometry(SubscribeOdometryRequest) returns(stream OdometryResponse) {}
 }
 
@@ -86,7 +89,7 @@ message AttitudeEulerResponse {
 
 message SubscribeAttitudeAngularVelocityBodyRequest {}
 message AttitudeAngularVelocityBodyResponse {
-    AngularVelocityBody attitude_angular_velocity_body = 1;
+    AngularVelocityBody attitude_angular_velocity_body = 1; // The next angular velocity (rad/s)
 }
 
 message SubscribeCameraAttitudeQuaternionRequest {}
@@ -306,18 +309,12 @@ message Odometry {
     uint64 time_usec = 1; // Timestamp (0 to use Backend timestamp).
     MavFrame frame_id = 2; // Coordinate frame of reference for the pose data.
     MavFrame child_frame_id = 3;  // Coordinate frame of reference for the velocity in free space (twist) data.
-    float x = 4; // X Position.
-    float y = 5; // Y Position.
-    float z = 6; // Z Position.
-    Quaternion q = 7; // Quaternion components, w, x, y, z (1 0 0 0 is the null-rotation).
-    float vx = 8; // X linear speed (m/s).
-    float vy = 9; // Y linear speed (m/s).
-    float vz = 10; // Z linear speed (m/s).
-    float rollspeed = 11; // Roll angular speed (rad/s).
-    float pitchspeed = 12; // Pitch angular speed (rad/s).
-    float yawspeed = 13; // Yaw angular speed (rad/s).
-    Covariance pose_covariance = 14; // Pose cross-covariance matrix.
-    Covariance velocity_covariance = 15; // Velocity cross-covariance matrix.
+    PositionXyz position = 4; // Position.
+    Quaternion q = 5; // Quaternion components, w, x, y, z (1 0 0 0 is the null-rotation).
+    SpeedBody speed_body = 6; // Linear speed (m/s).
+    AngularVelocityBody angular_velocity_body = 7; // Angular speed (rad/s).
+    Covariance pose_covariance = 8; // Pose cross-covariance matrix.
+    Covariance velocity_covariance = 9; // Velocity cross-covariance matrix.
 }
 
 /* Covariance type.
@@ -326,5 +323,19 @@ message Odometry {
  * Set first to NaN if unknown.
  */
 message Covariance {
-    repeated float c = 1;
+    repeated float covariance_matrix = 1;
+}
+
+// Speed type, represented in the Body (X Y Z) frame and in metres/second.
+message SpeedBody {
+    float velocity_x_m_s = 1; // Velocity in X in metres/second
+    float velocity_y_m_s = 2; // Velocity in Y in metres/second
+    float velocity_z_m_s = 3; // Velocity in Z in metres/second
+}
+
+// Position type, represented in the Body (X Y Z) frame
+message PositionXyz {
+    float x = 4; // X Position.
+    float y = 5; // Y Position.
+    float z = 6; // Z Position.
 }
--- a/protos/telemetry/telemetry.proto
+++ b/protos/telemetry/telemetry.proto
@@ -301,15 +301,15 @@ message Odometry {
     // Mavlink frame id
     enum MavFrame {
         UNDEF = 0;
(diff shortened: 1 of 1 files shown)


JonasVautherin:
I have no clue when that's required, so let me ask: does it make sense to have that in the `Telemetry` plugin, or could it be part of the `Mocap` one?

irsdkv:
> I have no clue when that's required, so let me ask: does it make sense to have that in the `Telemetry` plugin, or could it be part of the `Mocap` one?

I think that these should be placed in Telemetry, because of this can be used for various types of tasks. Such as estimators debug, testing Onboard Vision, or OTA visualization for example.

irsdkv:
Also it can be a good replacement for using a few other Telemetry messages (ATTITUDE_QUATERNION and ATTITUDE for example).

JonasVautherin:
> I think that it should be placed in Telemetry, because of this can be used for various types of tasks.

But is that a message that comes with every mavlink drone? Or does it exist when there is a mocap system? Can I just take any SITL model and use this message?

irsdkv:
> Can I just take any SITL model and use this message?

Yes. PX4 have "ODOMETRY" stream and it does not related to MoCap or onboard Vision.

protos/telemetry/telemetry.proto
```diff
@@ -286,3 +292,39 @@ enum LandedState{
...
+    }
+
+    uint64 time_usec = 1; // Timestamp (0 to use Backend timestamp).
+    MavFrame frame_id = 2; // Coordinate frame of reference for the pose data.
+    MavFrame child_frame_id = 3;  // Coordinate frame of reference for the velocity in free space (twist) data.
+    float x = 4; // X Position.
+    float y = 5; // Y Position.
+    float z = 6; // Z Position.
+    Quaternion q = 7; // Quaternion components, w, x, y, z (1 0 0 0 is the null-rotation).
+    float vx = 8; // X linear speed (m/s).
```
JonasVautherin: Would that be `SpeedBody`? We have `SpeedNed` already, it could make sense...

protos/telemetry/telemetry.proto
```diff
@@ -286,3 +292,39 @@ enum LandedState{
...
+    float vy = 9; // Y linear speed (m/s).
+    float vz = 10; // Z linear speed (m/s).
+    float rollspeed = 11; // Roll angular speed (rad/s).
```
JonasVautherin: Isn't that the existing `AngularVelocityBody`?

protos/telemetry/telemetry.proto
```diff
@@ -286,3 +292,39 @@ enum LandedState{
...
+        UNDEF = 0;
+        MAV_FRAME_BODY_NED = 8; // Setpoint in body NED frame. This makes sense if all position control is externalized - e.g. useful to command 2 m/s^2 acceleration to the right.
+        MAV_FRAME_VISION_NED = 16; // Odometry local coordinate frame of data given by a vision estimation system, Z-down (x: north, y: east, z: down).
+        MAV_FRAME_ESTIM_NED = 18; // Odometry local coordinate frame of data given by an estimator running onboard the vehicle, Z-down (x: north, y: east, z: down).
+    }
+
+    uint64 time_usec = 1; // Timestamp (0 to use Backend timestamp).
+    MavFrame frame_id = 2; // Coordinate frame of reference for the pose data.
+    MavFrame child_frame_id = 3;  // Coordinate frame of reference for the velocity in free space (twist) data.
+    float x = 4; // X Position.
```
JonasVautherin: Maybe in a `PositionXyz` message? Not sure how to call it...

irsdkv: I had the same difficulty :)

protos/telemetry/telemetry.proto
```diff
@@ -286,3 +295,47 @@ enum LandedState{
...
+
+// Speed type, represented in the Body (X Y Z) frame and in metres/second.
+message SpeedBody {
+    float velocity_x_m_s = 1; // Velocity in X in metres/second
+    float velocity_y_m_s = 2; // Velocity in Y in metres/second
+    float velocity_z_m_s = 3; // Velocity in Z in metres/second
+}
+
+// Position type, represented in the Body (X Y Z) frame
+message PositionXyz {
```
irsdkv: @JonasVautherin what about `PositionBody`? Like `SpeedBody` and `AngularVelocityBody`

protos/telemetry/telemetry.proto
```diff
@@ -286,3 +295,47 @@ enum LandedState{
     LANDED_STATE_TAKING_OFF = 3;
     LANDED_STATE_LANDING = 4;
 }
+
+// Odometry message type.
+message Odometry {
+    // Mavlink frame id
+    enum MavFrame {
+        UNDEF = 0;
+        MAV_FRAME_BODY_NED = 8; // Setpoint in body NED frame. This makes sense if all position control is externalized - e.g. useful to command 2 m/s^2 acceleration to the right.
```
julianoes: I wonder if we should expose the `MAV_FRAME` name or just use `BODY_NED`. @JonasVautherin opinions?

protos/telemetry/telemetry.proto
julianoes: I vote `PositionBody`.

protos/telemetry/telemetry.proto
```diff
@@ -286,3 +295,47 @@ enum LandedState{
...
+// Speed type, represented in the Body (X Y Z) frame and in metres/second.
+message SpeedBody {
+    float velocity_x_m_s = 1; // Velocity in X in metres/second
+    float velocity_y_m_s = 2; // Velocity in Y in metres/second
+    float velocity_z_m_s = 3; // Velocity in Z in metres/second
+}
+
+// Position type, represented in the Body (X Y Z) frame
+message PositionXyz {
+    float x = 4; // X Position.
```
julianoes: And here it would have to be `x_m`, I think.

protos/telemetry/telemetry.proto
JonasVautherin: You're right in that `BODY_NED` should be understandable. It's following the same naming as `PositionBody` etc.

`BODY_NED` is good for me :+1:.

protos/telemetry/telemetry.proto
```diff
@@ -286,3 +295,47 @@ enum LandedState{
     LANDED_STATE_TAKING_OFF = 3;
     LANDED_STATE_LANDING = 4;
 }
+
+// Odometry message type.
+message Odometry {
+    // Mavlink frame id
+    enum MavFrame {
```
julianoes: Sorry, I keep on being annoying here. Should we call it `Frame` instead of `MavFrame`?

protos/telemetry/telemetry.proto
```diff
@@ -286,3 +295,47 @@ enum LandedState{
...
+        UNDEF = 0;
+        BODY_NED = 8; // Setpoint in body NED frame. This makes sense if all position control is externalized - e.g. useful to command 2 m/s^2 acceleration to the right.
+        VISION_NED = 16; // Odometry local coordinate frame of data given by a vision estimation system, Z-down (x: north, y: east, z: down).
+        ESTIM_NED = 18; // Odometry local coordinate frame of data given by an estimator running onboard the vehicle, Z-down (x: north, y: east, z: down).
+    }
+
+    uint64 time_usec = 1; // Timestamp (0 to use Backend timestamp).
+    MavFrame frame_id = 2; // Coordinate frame of reference for the pose data.
+    MavFrame child_frame_id = 3;  // Coordinate frame of reference for the velocity in free space (twist) data.
+    PositionBody position = 4; // Position.
```
julianoes: What about `position_body`?

protos/telemetry/telemetry.proto
```diff
@@ -286,3 +295,47 @@ enum LandedState{
...
+// Speed type, represented in the Body (X Y Z) frame and in metres/second.
+message SpeedBody {
+    float velocity_x_m_s = 1; // Velocity in X in metres/second
+    float velocity_y_m_s = 2; // Velocity in Y in metres/second
+    float velocity_z_m_s = 3; // Velocity in Z in metres/second
+}
+
+// Position type, represented in the Body (X Y Z) frame
+message PositionBody {
+    float x_m = 4; // X Position in metres.
```
julianoes: Shouldn't this start at 1?

irsdkv: My bad. Fixed.

protos/telemetry/telemetry.proto
irsdkv: Hm. I think that `Frame` can be little confusing because of `MavFrame` (`MAV_FRAME`) is established term in docs and code

julianoes: Ok.

irsdkv:
@JonasVautherin @julianoes thank you for comments and suggestions
Do I need to do some more refactor or modifications?
I want to change `Odometry` struct in [Mocap](https://github.com/mavlink/MAVSDK-Proto/pull/105) plugin due to this `Odometry` message.

julianoes:
@irsdkv yes please adapt mocap.proto accordingly.

PR #111, opened 2019-11-22
telemetry: fix landed state enum

julianoes:
We don't need to add additional namespacing, I hope!


julianoes:
Ah, this was why:
```
protos/telemetry/telemetry.proto:292:5:"UNKNOWN" is already defined in "mavsdk.rpc.telemetry".
```

JonasVautherin:
Yes, unfortunately. And protobuf in Swift actually expects this namespacing, and removes the prefix when generating code. Of course that's not consistent with the other language implementations of protobuf, because that would be too easy :-).

PR #112, opened 2019-12-19
Added NOT_COMPLETED Result enum member in Shell proto

irsdkv:
Due to https://github.com/mavlink/MAVSDK/pull/944

### DIFF
--- a/protos/shell/shell.proto
+++ b/protos/shell/shell.proto
@@ -35,7 +35,8 @@ message ShellResult {
         NO_SYSTEM = 2; // No system is connected
         CONNECTION_ERROR = 3; // Connection error
         NO_RESPONSE = 4; // Response was not received
-        BUSY = 5; // Shell busy (transfer in progress)
+        NOT_COMPLETED = 5; // Response transfer potentially not completed (termination symbol not received)
+        BUSY = 6; // Shell busy (transfer in progress)
     }
 
     Result result = 1; // Result enum value


protos/shell/shell.proto
```diff
@@ -35,7 +35,8 @@ message ShellResult {
         NO_SYSTEM = 2; // No system is connected
         CONNECTION_ERROR = 3; // Connection error
         NO_RESPONSE = 4; // Response was not received
-        BUSY = 5; // Shell busy (transfer in progress)
+        NOT_COMPLETED = 5; // Response transfer potentially not completed (termination symbol not received)
```
JonasVautherin: Wait, now I'm confused :sweat_smile:. What's the difference between BUSY and NOT_COMPLETED? As a user, do I ever want to differentiate them?

irsdkv: Hmm. Yes, this really can confuse.
- `BUSY` - shell is busy now
- `NOT_COMPLETED` - in response of current request was not received `ESC`, so we suppose that transfer was interrupted .

May be need to use more clear name (or may be refuse this enum value)?

JonasVautherin: I think that it is the same thing for a user: the request has not completed, so a new request cannot be sent. I can't think of a case where a user says "if it's BUSY, do this, and if it's NOT_COMPLETED, do that". If there is no use-case for splitting them, I would vote for reusing `BUSY` (if you guys agree, of course).

irsdkv: Hmm. I think this is not the same. But now I think that `NOT_FINISHED` will be more likely to mislead the user than to help. 
May be only add some explaining to the `**brief**`? To notice user to try to increase delay time if response looks incomplete

JonasVautherin: ok :slightly_smiling_face:

irsdkv:
Hm. I thought we gave it up :smile:

JonasVautherin:
Not sure I understand that? You were saying that we need to have both `BUSY` and `NOT_COMPLETED`, because the user has to behave differently depending on which one is triggering, right?

irsdkv:
No, sorry for my English :)
I mean that `NOT_COMPLETED` is a redundant case. And talk about leave `BUSY` only and add notice to brief of [shell_command](https://github.com/mavlink/MAVSDK/blob/9895528bf81f687526d1360f06affa2d77e27938/src/plugins/shell/include/plugins/shell/shell.h#L78)

JonasVautherin:
Oh, my mistake, then. If you agree on keeping only `BUSY`, then I'm all for it. Sorry for merging too fast :sweat_smile:.

irsdkv:
Okay, I will reset corresponding commits in  MAVSDK PR and close MAVSDK-Python PR :)

JonasVautherin:
So we should revert this one as well, correct?

irsdkv:
Yes, it is superfluous now

JonasVautherin:
Ok I reverted it. Sorry for the confusion :sweat_smile:

PR #122, opened 2020-04-07
Add method options and field options

JonasVautherin:
For now, only MethodOptions is used to decide if a 'call' or 'request' method is 'async', 'sync' or 'both'.

For instance, set the PauseMission method to be 'sync' only with:

```
rpc PauseMission(PauseMissionRequest) returns(PauseMissionResponse) { option (mavsdk.options.async_type) = SYNC; }
```

FieldOptions are unused for now but can be used with e.g.:

```
double latitude_deg = 1 [(mavsdk.options.default_value)="NaN"]; // Latitude in degrees (range: -90 to +90)
```


JonasVautherin:
> we should add to the readme or docs on how to install this script.

Got it, will do :+1:

PR #136, opened 2020-04-23
shell: autogenerate C++ plugin, change API

julianoes:
The existing API with sync and async command requests did not work for commands with continuous outputs like top. Therefore we changed this to a simple async stream of output plus a sync request to send a command line.

### DIFF
--- a/protos/shell/shell.proto
+++ b/protos/shell/shell.proto
@@ -2,6 +2,8 @@ syntax = "proto3";
 
 package mavsdk.rpc.shell;
 
+import "mavsdk_options.proto";
+
 option java_package = "io.mavsdk.shell";
 option java_outer_classname = "ShellProto";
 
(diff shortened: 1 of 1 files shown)


protos/shell/shell.proto
```diff
@@ -2,40 +2,49 @@ syntax = "proto3";
...
 /**
  * Allow to communicate with the vehicle's system shell.
  */
 service ShellService {
-    // Communicate with a vehicle's Shell.
-    rpc Send(SendRequest) returns(SendResponse) {}
+    /*
+     * Send a command line.
+     */
+    rpc Send(SendRequest) returns(SendResponse) { option (mavsdk.options.async_type) = SYNC; }
```
JonasVautherin, line 17: I don't want to be annoying, so that's just out of curiosity. Why not combining both?

```rpc Send(SendRequest) returns(stream SendResponse) {}```

Because with them separate, there is no way to know which "ReceiveResponse" corresponds to which "SendRequest", is there?

julianoes: But you don't know when it ends, unless you wait for a timeout.

JonasVautherin: Oh, I thought they ended with `ESC`, or were infinite (like `top`), in which case we would need a way to cancel the call (which is already an issue now).

But if that's not a thing, then this API is fine. Thanks!

julianoes: Cancelling happens with 3x Enter, `q`, or Ctrl+C but that one is usually caught by your program already.

PR #137, opened 2020-04-23
Add proto file for MissionRaw

julianoes:
MissionRaw is coming to all the languages!

FYI @sanderux

PR #138, opened 2020-04-24
Autogenerate mocap plugin.

julianoes:
Contains #137 for now.

### DIFF
--- a/protos/mocap/mocap.proto
+++ b/protos/mocap/mocap.proto
@@ -8,7 +8,7 @@ option java_package = "io.mavsdk.mocap";
 option java_outer_classname = "MocapProto";
 
 /**
- * This class allows interfacing a vehicle with a motion capture system in
+ * Allows interfacing a vehicle with a motion capture system in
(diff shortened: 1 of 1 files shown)


protos/mocap/mocap.proto
```diff
@@ -2,78 +2,46 @@ syntax = "proto3";
...
 
+import "mavsdk_options.proto";
+
 option java_package = "io.mavsdk.mocap";
 option java_outer_classname = "MocapProto";
 
 /**
- * Motion Capture allow vehicles to navigate when a global
- * position source is unavailable or unreliable
+ * This class allows interfacing a vehicle with a motion capture system in
```
JonasVautherin, line 11:
```suggestion
 * Allows interfacing a vehicle with a motion capture system in
```

Nitpick: I don't think we need to say "this class", and I'm not sure if that would apply to any kind of bindings.

julianoes: Let me fix it!

julianoes:
Let me force push and check.

PR #144, opened 2020-04-29
Update follow me plugin for C++ autogeneration

julianoes:


protos/follow_me/follow_me.proto
```diff
@@ -11,63 +13,63 @@ option java_outer_classname = "FollowMeProto";
...
 
 // Target location for the vehicle to follow
 message TargetLocation {
-    double latitude_deg = 1; // Target latitude in degrees
-    double longitude_deg = 2; // Target longitude in degrees
-    double absolute_altitude_m = 3; // Target altitude in meters above MSL
-    float velocity_x_m_s = 4; // Target velocity in X axis, in meters per second
-    float velocity_y_m_s = 5; // Target velocity in Y axis, in meters per second
-    float velocity_z_m_s = 6; // Target velocity in Z axis, in meters per second
+    double latitude_deg = 1 [(mavsdk.options.default_value)="NAN"]; // Target latitude in degrees
```
JonasVautherin, line 50: I think it's `NaN` (which maybe is bad, I could change it) :thinking:

julianoes: Ok fixed.

PR #146, opened 2020-05-05
Add log_files plugin to auto-generation

julianoes:


protos/log_files/log_files.proto
```diff
@@ -0,0 +1,59 @@
...
+    repeated Entry entries = 2; // List of entries
+}
+
+message SubscribeDownloadLogFileRequest {
+    uint32 id = 1; // ID of the log file to download
+    string path = 2; // Path of where to download log file to.
+}
+message DownloadLogFileResponse {
+    LogFileResult log_file_result = 1;
+    float progress = 2; // Progress if result is progress
```
JonasVautherin, line 34: This assumes that we will never return anything other than this `float`. Fine for me, but just to make sure :-)

julianoes: I would think so, unless download speed :thinking: ...

julianoes: I'm changing it to a `ProgressData`.

julianoes:
Thx.

PR #150, opened 2020-05-08
minor fixes for MAVSDK-Java autogen

JonasVautherin:
* Java has different package names, so `autogen.py` needs to take that into account when ignoring files.
* The result type is hardcoded in Java, so it must be `<plugin_name.upper_camel_case>Result` in the proto file as well.
* The return name (here was `log_file_result` instead of `log_files_result`) should not require that, but it's faster to fix the Java autogen like this, and it doesn't hurt in the proto files.
* The languages bindings assume that a stream finishes on `SUCCESS`, gets an element on `NEXT`, and errors on everything else. Hence, `PROGRESS` is considered an error and should really be called `NEXT`.

Note that these changes should not have any influence in the compatibility between `mavsdk_server` and the language bindings: it's really just naming, but on the proto transport it uses the numbers, where `RESULT_NEXT = 2` and `RESULT_PROGRESS = 2` are encoded the same way. So these changes are not urgent and can come in the next `v0.25.1` release of `mavsdk_server`.

### DIFF
--- a/protos/log_files/log_files.proto
+++ b/protos/log_files/log_files.proto
@@ -21,7 +21,7 @@ service LogFilesService {
 
 message GetEntriesRequest {}
 message GetEntriesResponse {
-    LogFileResult log_file_result = 1;
+    LogFilesResult log_files_result = 1;
     repeated Entry entries = 2; // List of entries
 }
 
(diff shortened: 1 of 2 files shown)


protos/log_files/log_files.proto
```diff
@@ -21,7 +21,7 @@ service LogFilesService {
 
 message GetEntriesRequest {}
 message GetEntriesResponse {
-    LogFileResult log_file_result = 1;
+    LogFilesResult log_files_result = 1;
```
julianoes, line 24: Oh, that was just wrong.

PR #153, opened 2020-05-27
Add wifi plugin

JonasVautherin:
Not ready to merge yet, but open for comments


JonasVautherin:
I'm not super convinced by the naming: `GetAccessPointConfiguration` can say `ACCESS_POINT` or `STATION`. Should it be just `GetConfiguration`?

Also it's a draft because `Mode` will clash with the camera plugin. I need to find a way in the Python generated code.

julianoes:
I'd say `GetConfiguration` is better.

PR #154, opened 2020-05-29
Adding new plugin for manual control

AldoCF96:
Adding a sketch for the plugin. 
Adding new features to control the drone in the X,Y,Z and R position
Adding features like connect, arming,disarming, setting the drone in position mode.


protos/joystick/joystick.proto
```diff
@@ -0,0 +1,138 @@
+syntax = "proto3";
+
+package mavsdk.rpc.joystick;
```
julianoes, line 3: I would probably call it `manual_control` because it is not limited to joystick. It could also be gamepad, etc.

And you could also use a joystick with offboard control.

protos/joystick/joystick.proto
```diff
@@ -0,0 +1,138 @@
...
+
+option java_package = "io.mavsdk.joystick";
+option java_outer_classname = "JoystickProto";
+
+service JoystickService
+{
+
+    rpc ConnectDrone (ConnectRequest)  returns (ConnectResponse) {}
```
julianoes, line 11: I don't think this `ConnectRequest` is required.

protos/joystick/joystick.proto
```diff
@@ -0,0 +1,138 @@
...
+    /*
+     * Send command to enable Joystick mode 
+     *
+     * Enabling Joystic mode causes to change into Position Control mode
+     * drone still needs to be armed
+     * Needs to check if the drone is not currently flying or in mission if so drone needs to be paused first 
+     */
+    
+    rpc EnableJoystickMode (EnableRequest)  returns(EnableResponse) {}
```
julianoes, line 20: We should probably make this to set a flight mode e.g. Altitude Control or Position Control.

protos/joystick/joystick.proto
```diff
@@ -0,0 +1,138 @@
...
+
+
+
+    rpc ArmDrone (ArmRequest) returns (ArmResponse) {}
+
+
+    rpc DisarmDrone (DisarmRequest) returns (DisarmResponse) {}
```
julianoes, line 27: We can re-use `Arm` and `Disarm` from the action plugin.

protos/joystick/joystick.proto
```diff
@@ -0,0 +1,138 @@
...
+
+     /*
+     * Send command to command to the drone to move up  or down 
+     */
+    rpc Zmotion (ZmotionRequest) returns (ZmotionResponse) {}
+
+     /*
+     * Send command to command to rotate clockwise or counterclockwise 
+     */
+    rpc Rmotion (RmotionRequest) returns (RmotionResponse) {}
```
julianoes, line 47: Do you really need specific requests for each axis or could it be one request containing all 4 axes at once?

protos/joystick/joystick.proto
AldoCF96: I think it is needed, lets say your flying more than 1 drone, then which system do we choose ?

protos/joystick/joystick.proto
AldoCF96: Do I delete this or keep it ?

protos/joystick/joystick.proto
AldoCF96: Assuming 2 joystick sticks, each stick will control 2 axis. 
I think it can be done all at once,  I just wanted a way to assign the axis to each stick, but I can see that it is better to put all of then at once.

protos/joystick/joystick.proto
AldoCF96: ok will do

protos/joystick/joystick.proto
JonasVautherin: Multi-drone support is something that needs to be solved at the MAVSDK level, not at a plugin level. So I would agree with Julian here :blush:

protos/joystick/joystick.proto
JonasVautherin: I would vote for deleting. No need to duplicate the features, and those ones exist in the `action` plugin.

protos/joystick/joystick.proto
JonasVautherin: I am not super experienced with the [`MANUAL_CONTROL`](https://mavlink.io/en/messages/common.html#MANUAL_CONTROL) message, but when I used it for tests, I was sending one message with all the values, at 20Hz. At the MAVLink implementation level, I think we will need to aggregate those values into one message anyway, right? Otherwise the values that are not set will be interpreted at "0", won't they?

One question, though, is whether it should work like the offboard plugin (i.e. you set only the change of position of the joystick, and MAVSDK keeps the stick in that position until you actively change it), or if we should delegate the sending of those messages to the user (i.e. the MAVSDK user must send the message at some rate, e.g. 10-20Hz). Opinions there?

AldoCF96: I have push the new changes, also changing the name to manual control so it makes more sense

protos/manual_control/manual_control.proto
```diff
@@ -0,0 +1,112 @@
...
+
+
+message MotionRequest {
+    sint32 xmotion = 1 ;                    // value between -1000 to 1000. negative-> backwards, positive-> forwards 
+    sint32 ymotion = 2 ;                    // value between -1000 to 1000. negative-> left, positive--> right
+    sint32 zmotion = 3 ;                    // value between -1000 to 1000. negative-> down ,positive--> up 
+    sint32 revolve = 4 ;                    // value between -1000 to 1000. negative-> ccw ,positive- -> cw  revolve aroun z-axis 
+}
+message MotionResponse{
+    ManualControlResult  mc_response = 1  ;
```
JonasVautherin, line 83: Because of how auto-generation works, the `ManualControlResult` has to be called `manual_control_result`:

```suggestion
    ManualControlResult  manual_control_result = 1  ;
```

protos/joystick/joystick.proto
JonasVautherin: Thinking about my point above, I think that the user will anyway have to repeatedly send the value when it changes, so we should delegate that (unlike the offboard plugin).

protos/manual_control/manual_control.proto
AldoCF96: done

protos/joystick/joystick.proto
AldoCF96: I see will do

protos/joystick/joystick.proto
AldoCF96: I think it should a neutral position(hold position) in this case that is both sticks on the centre, and the message should be send as long it is move from the initial position, so when the initial position still there should not be a message sent, otherwise  move the sticks and hold them at position to obtain motion.  to stop motion sticks should be released. 

But you are right messages should be send when the stick changes position however, there should be a threshold to use +-10 depending on the input I do not know about the sensibility of the joystick and how those values are read.  there is not really a point on sending the same message 10 to 20 times in a second if they are all the same. 

should it be a state that detects that a joystick is active and if it is inactive it triggers a safemode ?

JonasVautherin: > there is not really a point on sending the same message 10 to 20 times in a second if they are all the same

Are you sure about that? Do you already use `MANUAL_CONTROL` in your code? In my experience, the message has to be sent repeatedly no matter what, otherwise PX4 reports "data link lost" or something like that.

protos/joystick/joystick.proto
AldoCF96: Ok I will delete it

protos/joystick/joystick.proto
AldoCF96: I  havent actually implemented  it. thanking for letting me know

protos/manual_control/manual_control.proto
```diff
@@ -0,0 +1,85 @@
...
+    
+    /*
+     * Send command to enable Joystick mode 
+     *
+     * Enabling Joystic mode causes to change into Position Control mode
+     * drone still needs to be armed
+     * Needs to check if the drone is not currently flying or in mission if so drone needs to be paused first 
+     */
+    
+    rpc SetPositionControl (PositionControlRequest)  returns(PositionControlResponse) {}
```
julianoes, line 20: I think we should add either:
- something like `action->SetFlightMode(FlightMode::PositionControl)`
- or call it `StartPositionControl` becaue it' not a "setter" as it's not setting anything.

And the comment should only describe the API and not the internal implementation, so you can keep it simple.

protos/manual_control/manual_control.proto
```diff
@@ -0,0 +1,85 @@
...
+    
+    
+    rpc SetAltitudeControl (AltitudeControlRequest)  returns(AltitudeControlResponse) {}
+    
+      /*
+     * Send command to command to the drone to move forwards or backwards 
+     */
+    rpc Axismotion (MotionRequest) returns (MotionResponse) {}
```
julianoes, line 28: Axismotion is a term I don't know. What about `ManualControlInput` or `ManualSetpoint`?

protos/manual_control/manual_control.proto
```diff
@@ -0,0 +1,85 @@
...
+
+message AltitudeControlRequest{}
+message AltitudeControlResponse{
+    ManualControlResult manual_control_result = 1;
+}
+
+
+
+message MotionRequest {
+    sint32 xmotion = 1 ;                    // value between -1000 to 1000. negative-> backwards, positive-> forwards
```
julianoes, line 50:
```suggestion
    float32 x = 1 ;                    // value from -1 to 1. negative-> backwards, positive-> forwards 
```

I wouldn't constrain it to a resolution of 1000 but just use float.

protos/manual_control/manual_control.proto
JonasVautherin: If MAVLink does not specify which mode goes with the `MANUAL_CONTROL` message (and hence the user has to know it), then I think I would vote for `action->SetFlightMode`... and that would centralize it for when MAVLink flight modes get better specified. Does that make sense?

protos/manual_control/manual_control.proto
```diff
@@ -0,0 +1,85 @@
+syntax = "proto3";
+
+package mavsdk.rpc.manual_control;
+
+option java_package = "io.mavsdk.manual_control";
+option java_outer_classname = "Manual_ControlProto";
```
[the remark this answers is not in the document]
AldoCF96: will change it

protos/manual_control/manual_control.proto
AldoCF96: I did it base on the messages documentation which varies betweem -1000 to 1000

JonasVautherin: Yes that's a good point.

@julianoes: I'm not completely sure I understand why a float is better, given that the mavlink message has a fixed resolution. My concern is that I could imagine ending up in situations where the user sends a float, and receives back another float but both were represented by the same MAVLink message.

protos/manual_control/manual_control.proto
julianoes: Either remove it and then we need to add a flight mode setter to `Action`, or rename it to `StartPositionControl` as suggested above.

protos/manual_control/manual_control.proto
julianoes: I see. I generally advocate not to be guided by the MAVLink API but try to do what is right, makes sense. And it doesn't really make sense to put values from -1000 to 1000 into a type which allows for -32767 to 32768.

If we really do make it from -1000 to 1000 we should use the same type as the MAVLink message as well which is `int16`.

> @JonasVautherin
> and receives back another float but both were represented by the same MAVLink message

How would that happen exactly?

AldoCF96: I am not sure that int16 exis on the proto library that is the reason I  use sint32

protos/manual_control/manual_control.proto
```diff
@@ -0,0 +1,85 @@
...
+    
+    rpc StartPositionControl (PositionControlRequest)  returns(PositionControlResponse) {}
+    
+    
+    rpc StartAltitudeControl (AltitudeControlRequest)  returns(AltitudeControlResponse) {}
+    
+      /*
+     * Send command to command to the drone to move forwards or backwards 
+     */
+    rpc ManualControlInput (MotionRequest) returns (MotionResponse) {}
```
JonasVautherin, line 28:
```suggestion
    rpc ManualControlInput (ManualControlRequest) returns (ManualControlResponse) {}
```

protos/manual_control/manual_control.proto
```diff
@@ -0,0 +1,85 @@
...
+    
+    /*
+     * Send command to enable Joystick mode 
+     *
+     * Enabling Joystic mode causes to change into Position Control mode
+     * drone still needs to be armed
+     * Needs to check if the drone is not currently flying or in mission if so drone needs to be paused first 
+     */
+    
+    rpc StartPositionControl (PositionControlRequest)  returns(PositionControlResponse) {}
```
JonasVautherin, line 20: @julianoes: Didn't we say that setting the flight mode should be an action?

julianoes: Either that or this at least for now.

I'm reluctant to add a flight mode setter before we have proper MAVLink flight modes.

julianoes:
> I am still having troubles about the C++ files generation. Also how do I proceed from there ?

What troubles? Can we help?

protos/manual_control/manual_control.proto
AldoCF96: should I keep it or delete it ?

JonasVautherin: Let's keep it for now

JonasVautherin:
Nitpick: @AldoCF96 would you mind removing the useless empty lines a bit everywhere in the file? So that it is consistent with the other proto files :innocent:.

AldoCF96:
> > I am still having troubles about the C++ files generation. Also how do I proceed from there ?
> 
> What troubles? Can we help?

Yeah, I havent been able to generate the C++ files, is the tutorial for adding new plugins done. I am not sure how to proceed?

AldoCF96:
> Nitpick: @AldoCF96 would you mind removing the useless empty lines a bit everywhere in the file? So that it is consistent with the other proto files .

sure will do

AldoCF96:
> > I am still having troubles about the C++ files generation. Also how do I proceed from there ?
> 
> What troubles? Can we help?
I am having this trouble
image

julianoes:
Ok, looks like something changed with the shared lib `libgrpc_plugin_support.so.1` that you need. Try to rebuild your `grpc_cpp_plugin` by doing:
```
cmake -DBUILD_BACKEND=ON -Bbuild/default -H.
```

AldoCF96:
> cmake -DBUILD_BACKEND=ON -Bbuild/default -H.

I tried, it stills shows the same error. I am not sure how to fix that

protos/manual_control/manual_control.proto
```diff
@@ -0,0 +1,63 @@
...
+option java_package = "io.mavsdk.manual_control";
+option java_outer_classname = "ManualControlProto";
+
+service ManualControlService
+{
+     /*
+     * Send command to enable Joystick mode 
+     * Enabling Joystic mode causes to change into Position Control mode
+     * drone still needs to be armed
+     * Needs to check if the drone is not currently flying or in mission if so drone needs to be paused first
```
julianoes, line 14:
```suggestion
     * Start position control using e.g. joystick input.
     * 
     * Requires a valid position using e.g. GPS, external vision, or optical flow.
```

protos/manual_control/manual_control.proto
```diff
@@ -0,0 +1,63 @@
...
+service ManualControlService
+{
+     /*
+     * Send command to enable Joystick mode 
+     * Enabling Joystic mode causes to change into Position Control mode
+     * drone still needs to be armed
+     * Needs to check if the drone is not currently flying or in mission if so drone needs to be paused first 
+     */
+    rpc StartPositionControl (PositionControlRequest)  returns(PositionControlResponse) {}
+    rpc StartAltitudeControl (AltitudeControlRequest)  returns(AltitudeControlResponse) {}
```
julianoes, line 17:
```suggestion
    /*
     * Start altitude control 
     *
     * Does not require a  valid position e.g. GPS.
     */
    rpc StartAltitudeControl (AltitudeControlRequest)  returns(AltitudeControlResponse) {}
```

protos/manual_control/manual_control.proto
```diff
@@ -0,0 +1,63 @@
...
+     /*
+     * Send command to command to the drone to move forwards or backwards 
+     */
+    rpc ManualControlInput (ManualControlRequest) returns (ManualControlResponse) {}
```
julianoes, line 21: This would  be stabilized mode, right? I'm not sure that we should support this because it's generally hard to fly stabilized unless you have a connection with quite low latency.

AldoCF96: is stabilized mode  same as position mode ?
how much means here low latency ?

julianoes: Stabilized is often called Manual but it's basically attitude control with direct throttle. The direct throttle part is what requires some skill and low latency.

I would, for now, not implement it, unless you actually require it. Would that make sense?

julianoes: Ok, I'm removing it for now.

julianoes: Oh, I was confused here actually. This is not to change to manual mode but actually setting the manual input :man_facepalming:. Ok I'll bring it back!

protos/manual_control/manual_control.proto
```diff
@@ -0,0 +1,69 @@
...
+    float z = 3; // value between -1. to 1. negative -> down, positive -> up (usually for now, for multicopter 0 to 1 is expected)
+    float r = 4; // value between -1. to 1. negative -> turn anti-clockwise (towards the left), positive -> turn clockwise (towards the right)
+}
+message SetManualControlInputResponse {
+    ManualControlResult  manual_control_result = 1;
+}
+
+message ManualControlResult {
+    enum Result {
+
```
JonasVautherin, line 57:
```suggestion
```

protos/manual_control/manual_control.proto
```diff
@@ -0,0 +1,69 @@
...
+        RESULT_UNKNOWN = 0; // Unknown result
+        RESULT_SUCCESS = 1; // Request was successful
+        RESULT_NO_SYSTEM = 2; // No system is connected
+        RESULT_CONNECTION_ERROR = 3; // Connection error
+        RESULT_BUSY = 4; // Vehicle is busy
+        RESULT_COMMAND_DENIED = 5; // Command refused by vehicle
+        RESULT_TIMEOUT = 6; // Request timed out
+        RESULT_INPUT_OUT_OF_RANGE = 7; // Input out of range
+    }
```
JonasVautherin, line 66:
```suggestion
    }

```

PR #164, opened 2020-07-09
Added support to reboot/shutdown seperate components

shakeebbb:
Pull request for the issue mavlink/MAVSDK#1137

### DIFF
--- a/protos/action/action.proto
+++ b/protos/action/action.proto
@@ -136,12 +136,16 @@ message LandResponse {
     ActionResult action_result = 1;
 }
 
-message RebootRequest {}
+message RebootRequest {
+    uint32 mav_component = 1;
+}
 message RebootResponse {
     ActionResult action_result = 1;
 }
 
-message ShutdownRequest {}
+message ShutdownRequest {
+    uint32 mav_component = 1;
+}
 message ShutdownResponse {
     ActionResult action_result = 1;
 }


protos/action/action.proto
```diff
@@ -136,12 +136,16 @@ message LandResponse {
     ActionResult action_result = 1;
 }
 
-message RebootRequest {}
+message RebootRequest {
+    uint32 mav_component = 1;
```
julianoes, line 140:
```suggestion
    uint8 mavlink_component_id = 1; // MAVLink component ID of component to reboot (0 for all).
```

protos/action/action.proto
```diff
@@ -136,12 +136,16 @@ message LandResponse {
...
+}
 message RebootResponse {
     ActionResult action_result = 1;
 }
 
-message ShutdownRequest {}
+message ShutdownRequest {
+    uint32 mav_component = 1;
```
julianoes, line 147:
```suggestion
    uint8 mavlink_component_id = 1; // MAVLink component ID of component to reboot (0 for all).
```

protos/action/action.proto
JonasVautherin: Sorry I commented [there](https://github.com/mavlink/MAVSDK/pull/1138#discussion_r451633446). I would not leak the mavlink compid here.

IMO `action.reboot()` talks to the autopilot. Because one may want to reboot everything or only the autopilot, maybe we could have a boolean, like `action.reboot(boolean autopilot_only)`, and by default it would just reboot everything. Then if one needs a way to reboot a camera or a gimbal, I would put that into `drone.camera.reboot()` or `drone.gimbal.reboot()`.

julianoes: Oh, right, I forgot about your suggestion. Ok so we change it to only reboot autopilot, so we don't need to change proto except for a commend saying "only reboots the autopilot"?

JonasVautherin: If I remember correctly, @LorenzMeier added that feature, and the intent was to reboot everything. Hence my suggestion of defaulting to everything, but adding a boolean to limit that to `autopilot_only`. So that we don't break backward compatibility :innocent:. Would that be an acceptable API in your opinion?

shakeebbb: When 'mavlink_component_id' is 0 or MAV_COMP_ID_ALL, it reboots everything just as it was doing before. So backward compatibility can be intact if we put the default 'mavlink_component_id' as 0.

JonasVautherin: Yes, at the MAVLink level. But the MAVSDK API does not have to expose that there is a "MAVLink component id" at all. We could say that `drone.action.reboot()` sends the reboot command with compid MAV_COMP_ID_ALL, and `drone.action.reboot(true)` (i.e. `autopilot_only = true`) would only reboot the autopilot. So that a user of MAVSDK does not have to know about the concept of component id. Does that make sense?

julianoes: @JonasVautherin ok what should we do?

JonasVautherin:
```suggestion
    bool autopilot_only = 1;
```

Because [bools default to 'false'](https://developers.google.com/protocol-buffers/docs/proto3#default) in protobuf, this will not break backward compatibility, but it will allow rebooting only the autopilot (when passing 'true'), which is the new use-case.

That would be my suggestion :innocent:.

protos/action/action.proto
JonasVautherin:
```suggestion
    bool autopilot_only = 1;
```

protos/action/action.proto
julianoes: ok sounds good.

JonasVautherin:
@shakeebbb are you clear on what needs to be done here? :innocent:

PR #166, opened 2020-07-16
New action: landing_fixed_wing

iwishiwasaneagle:
Request parameters based on QGC.
Implementation will follow [px4 fixed_wing_landing documentation](https://docs.px4.io/v1.9.0/en/flying/fixed_wing_landing.html).

### DIFF
--- a/protos/action/action.proto
+++ b/protos/action/action.proto
@@ -143,14 +143,14 @@ message LandResponse {
 }
 
 message LandFixedWingRequest {
-    float loiter_alt = 40.0; // Loiter altitude (in meters)
-    float loiter_rad = 75.0; // Loiter radius (in meters)
-    bool loiter_cw = true; // Loiter clockwise
-    float land_heading = 0.0; // Landing heading (in degrees)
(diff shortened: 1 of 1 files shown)


iwishiwasaneagle:
After further investigation, this actually doesn't seem feasible as 1) there's no native PX4 way to land a fixed-wing plane and 2) it's hard to encode all the data for a good landing in just a command

JonasVautherin:
Does that mean it has to be done in a mission?

iwishiwasaneagle:
> Does that mean it has to be done in a mission?

It has to be done via a mission as far as I can tell, you can see this here https://gist.github.com/iwishiwasaneagle/0c16e5c98e02c9b4eab1c7d83ff26a8d. 
Basically at least 2 MAV_CMDs are called -> MAV_CMD_DO_LAND_START and MAV_CMD_NAV_LAND. However something like MAV_CMD_NAV_LOITER_TO_ALT or a waypoint should be added to specify a glide slope of sorts for FW.

PR #167, opened 2020-07-16
New plugin: mission_fixed_wing

iwishiwasaneagle:
As discussed with @julianoes on the Slack, here's a first draft for `mission_fixed_wing` largely based on `mission.proto` with adjustments to make it suitable for fixed-wing UAVs.

### DIFF
--- a/protos/mission_fixed_wing/mission_fixed_wing.proto
+++ b/protos/mission_fixed_wing/mission_fixed_wing.proto
@@ -1,0 +1,243 @@
+syntax = "proto3";
+
+package mavsdk.rpc.mission_fixed_wing;
+
+import "mavsdk_options.proto";
+
+option java_package = "io.mavsdk.mission_fixed_wing";
+option java_outer_classname = "MissionFixedWingProto";
+
+// Enable waypoint missions.
+service MissionService {
+    /*
+     * Upload a list of mission items to the system.
+     *
+     * The mission items are uploaded to a fixed wing drone. Once uploaded the mission can be started and
+     * executed even if the connection is lost.
+     */
+    rpc UploadMission(UploadMissionRequest) returns(UploadMissionResponse) {}
+    /*
+     * Cancel an ongoing mission upload.
+     */
+    rpc CancelMissionUpload(CancelMissionUploadRequest) returns(CancelMissionUploadResponse) { option (mavsdk.options.async_type) = SYNC; }
+    /*
+     * Download a list of mission items from the system (asynchronous).
+     *
+     * Will fail if any of the downloaded mission items are not supported
+     * by the MAVSDK API.
+     */
+    rpc DownloadMission(DownloadMissionRequest) returns(DownloadMissionResponse) {}
+    /*
+     * Cancel an ongoing mission download.
+     */
+    rpc CancelMissionDownload(CancelMissionDownloadRequest) returns(CancelMissionDownloadResponse) { option (mavsdk.options.async_type) = SYNC; }
+    /*
+     * Start the mission.
+     *
+     * A mission must be uploaded to the vehicle before this can be called.
+     */
+    rpc StartMission(StartMissionRequest) returns(StartMissionResponse) {}
+    /*
+     * Pause the mission.
+     *
+     * Pausing the mission puts the vehicle into
+     * [HOLD mode](https://docs.px4.io/en/flight_modes/hold.html).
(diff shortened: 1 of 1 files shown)


protos/mission_fixed_wing/mission_fixed_wing.proto
```diff
@@ -0,0 +1,243 @@
...
+    double longitude_deg = 2 [(mavsdk.options.default_value)="NaN", (mavsdk.options.epsilon)=0.0000001]; // Longitude in degrees (range: -180 to +180)
+    float relative_altitude_m = 3 [(mavsdk.options.default_value)="NaN"]; // Altitude relative to takeoff altitude in metres
+    float speed_m_s = 4 [(mavsdk.options.default_value)="NaN"]; // Speed to use after this mission item (in metres/second)
+    bool is_fly_through = 5 [(mavsdk.options.default_value)="false"]; // True will make the fixed-wing drone fly through without stopping, while false will make the fixed-wing drone hold around the waypoint
+    float gimbal_pitch_deg = 6 [(mavsdk.options.default_value)="NaN"]; // Gimbal pitch (in degrees)
+    float gimbal_yaw_deg = 7 [(mavsdk.options.default_value)="NaN"]; // Gimbal yaw (in degrees)
+    CameraAction camera_action = 8; // Camera action to trigger at this mission item
+    float loiter_time_s = 9 [(mavsdk.options.default_value)="NaN"]; // Loiter time (in seconds)
+    double camera_photo_interval_s = 10 [(mavsdk.options.default_value)="1.0"]; // Camera photo interval to use after this mission item (in seconds)
```
julianoes, line 198: So we probably need an enum here for takeoff and landing, something like `FixedwingTakeoff`, `FixedwingApproach`, `FixedwingLanding`, right?

iwishiwasaneagle: Is clumping those 3 into the 1 `MissionItem` the right way of doing it?

julianoes: How else would you add it? I'm open for suggestions.

julianoes: We could have a MissionPlan consisting of a takeoff part, flight part, and land part, for instance...

iwishiwasaneagle: Well for a valid mission you need a landing part. Take off is only need if you're grounded. So maybe have it such that `optional takeoff creator` -> `mission planning as it currently is` -> `mandatory landing creator`.  

I was just tinkering about with how to construct a valid fixed-wing mission and this seems to be the bare useful minimum that's allowed by PX4. 

```python
#!/usr/bin/env python3

import asyncio
from mavsdk import System, Mission_raw

async def run():
    drone = System()
    await drone.connect(system_address="udp://:14540")

    print("Waiting for drone to connect...")
    async for state in drone.core.connection_state():
        if state.is_connected:
            print(f"Drone discovered with UUID: {state.uuid}")
            break
    # Landing with a 6deg slope calcualted by QGC
    items = []
    items.append(
        mission_raw.MissionItem( # MAV_CMD_DO_LAND_START 189
            0,2,189,1,1,0,0,0,0,0,0,0,0)
    )
    items.append( # MAV_CMD_NAV_LOITER_TO_ALT 31
        mission_raw.MissionItem(
            1,6,31,0,1,1,75,0,1,558683815,-42929235,40.0,0)
    )
    items.append( # MAV_CMD_NAV_LAND 21
        mission_raw.MissionItem(
            2,6,21,0,1,0,0,0,0,558688774,-42867694,0,0
        )
    )
    await drone.mission_raw.upload_mission(items)
    await drone.mission_raw.start_mission()

if __name__ == "__main__":
    loop = asyncio.get_event_loop()
    loop.run_until_complete(run())
```

JonasVautherin:
Thanks a lot @iwishiwasaneagle! We had discussions in the past with @hamishwillee and @julianoes about this architecture (i.e. having different proto files for the different platforms [fw, mc, vtol, rover, ...]), but I recently came to think that maybe it would be better to keep one proto file per plugin (that would have the functions for all the platforms), but the user would have to call the right function by testing the platform (with something like `Drone.isMulticopter()`).

I wrote it in more details here, please share your opinion :blush:: https://github.com/mavlink/MAVSDK-Proto/issues/168

PR #172, opened 2020-08-12
Add set actuator

coderkalyan:
Adds support for the new MAV_CMD_DO_SET_ACTUATOR mavlink command.

### DIFF
--- a/protos/action/action.proto
+++ b/protos/action/action.proto
@@ -75,6 +75,10 @@ service ActionService {
      */
     rpc GotoLocation(GotoLocationRequest) returns(GotoLocationResponse) {}
     /*
+     * Send command to set the value of an actuator.
+     */
+    rpc SetActuator(SetActuatorRequest) returns(SetActuatorResponse) {}
+    /*
      * Send command to transition the drone to fixedwing.
      *
      * The associated action will only be executed for VTOL vehicles (on other vehicle types the
@@ -163,6 +167,14 @@ message GotoLocationRequest {
     float yaw_deg = 4; // Yaw angle (in degrees, frame is NED, 0 is North, positive is clockwise)
 }
 message GotoLocationResponse {
+    ActionResult action_result = 1;
+}
+
+message SetActuatorRequest {
+    int32 index = 1; // Index of actuator (starting with 1)
+    float value = 2; // Value to set the actuator to (normalized from [-1..1])
+}
+message SetActuatorResponse {
     ActionResult action_result = 1;
 }
 


JonasVautherin:
@julianoes: Do you have an opinion on this? Not sure if that's an `action` or not :innocent:

coderkalyan:
@JonasVautherin would you rather I pull this out into a separate plugin? I'm planning to add a few more commands for various payload/actuator related stuff (winch, gripper, etc) so I'm wondering if putting them along with this in a new `Payload` or similar plugin would be a good idea?

As for the current plugins, `action` seemed the most appropriate IMO.

JonasVautherin:
Yeah I'm not completely sure. If there are more coming, maybe it makes sense to group them in a new plugin indeed. Let's wait for @julianoes' opinion maybe, he'll be back next week :slightly_smiling_face:

hamishwillee:
FWIW I tend to see most non-camera payloads as actions - parachute, gripper, winch etc as `Actions`, on the basis that this is where I would look for them. What do you see as the benefit to separating them? 

Either way, would not call them payload, as that will confuse the camera folks :-)

I also somewhat dislike having actuator/servo grouped with everything else. Why? Well I see servo and actuator commands as "test commands" before someone has done the proper work of adding support for a real payload command. Contrast with a  winch, gripper, whatever command - the user doesn't have to do anything other than say "work". these ones can be used in missions too, because they have a known and fixed meaning.

protos/action/action.proto
```diff
@@ -74,6 +74,12 @@ service ActionService {
      * The yaw angle is in degrees (frame is NED, 0 is North, positive is clockwise).
      */
     rpc GotoLocation(GotoLocationRequest) returns(GotoLocationResponse) {}
+    /*
+     * Send command to set the value of an actuator.
+     *
+     * The index of the actuator is 1-indexed, and the value is normalized from [-1..1]
```
julianoes, line 80: I would remove this line because it is already described below.

protos/action/action.proto
```diff
@@ -166,6 +172,14 @@ message GotoLocationResponse {
     ActionResult action_result = 1;
 }
 
+message SetActuatorRequest {
+  int32 index = 1; // Index of actuator (starting with 1)
```
julianoes, line 176: I assume that's 1-indexed because that's how the MAVLink message is specced?

julianoes (commented):
Thanks for that. See inline comments.

And @JonasVautherin I don't know what other plugin this should be in. Do you have an idea?

protos/action/action.proto
coderkalyan: Ok.

protos/action/action.proto
coderkalyan: Yep. Is this a problem?

coderkalyan:
Ok @julianoes I updated the docstring, should be good to go there. We still need to figure out if this should be an Action or not.

@hamishwillee I don't really like associating `Payload` with `Camera` since a camera is a camera and a payload in my mind is some sort of cargo or cargo-related action, but I see how that can confuse some people so I guess we might want to avoid that name.

protos/action/action.proto
julianoes: No, I guess it's ok.

JonasVautherin:
> I don't know what other plugin this should be in. Do you have an idea?

Nope I don't. I was just not sure if this was somehow lower-level than the usual actions we have. But I'm fine with having them as actions :+1:.

protos/action/action.proto
```diff
@@ -166,6 +170,14 @@ message GotoLocationResponse {
...
+  float value = 2; // Value to set the actuator to (normalized from [-1..1])
```
julianoes, line 175: Ha I found something to complain! :smile: 

Please match indentation to 4 spaces everywhere.

coderkalyan: Whoops, not sure how that happened. Will fix

coderkalyan:
@julianoes Fixed the indentation, if there's nothing else wrong I think we can merge this? Unless you want me to pull this into its own plugin, and I'm not sure if thats necessary at this point in time. But lmk.

FYI @JonasVautherin @hamishwillee if you have anything else to add about that, feel free.

julianoes (approved):

julianoes:
@coderkalyan one thing: it needs to be supported by PX4 first :smile:

coderkalyan:
@julianoes oh, ok. I'll let you know when my PX4 Firmware PR gets merged :smiley:

PR #176, opened 2020-10-06
#1207 extra fields for fixedwing metrics

haydockjp:
There are three extra fields in fixed wing metrics that are not returned.
- ground speed
- altitude
- heading

### DIFF
--- a/protos/telemetry/telemetry.proto
+++ b/protos/telemetry/telemetry.proto
@@ -601,6 +601,9 @@ message FixedwingMetrics {
     float airspeed_m_s = 1 [(mavsdk.options.default_value)="NaN"]; // Current indicated airspeed (IAS) in metres per second
     float throttle_percentage = 2 [(mavsdk.options.default_value)="NaN"]; // Current throttle setting (0 to 100)
     float climb_rate_m_s = 3 [(mavsdk.options.default_value)="NaN"]; // Current climb rate in metres per second
+    float groundspeed_m_s = 4 [(mavsdk.options.default_value)="NaN"]; // Current indicated groundspeed in metres per second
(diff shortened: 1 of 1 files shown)


protos/telemetry/telemetry.proto
```diff
@@ -601,6 +601,9 @@ message FixedwingMetrics {
     float airspeed_m_s = 1 [(mavsdk.options.default_value)="NaN"]; // Current indicated airspeed (IAS) in metres per second
     float throttle_percentage = 2 [(mavsdk.options.default_value)="NaN"]; // Current throttle setting (0 to 100)
     float climb_rate_m_s = 3 [(mavsdk.options.default_value)="NaN"]; // Current climb rate in metres per second
+    float groundspeed_m_s = 4 [(mavsdk.options.default_value)="NaN"]; // Current indicated groundspeed in metres per second
+    float alt_m = 5 [(mavsdk.options.default_value)="NaN"]; // Current altitude (MSL) in metres
```
julianoes, line 605:
```suggestion
    float altitude_m = 5 [(mavsdk.options.default_value)="NaN"]; // Current altitude (MSL) in metres
```

protos/telemetry/telemetry.proto
```diff
@@ -601,6 +601,9 @@ message FixedwingMetrics {
...
+    uint32 heading_deg = 6 [(mavsdk.options.default_value)="NaN"]; // Current heading in compass units (0-360, 0=north)
```
julianoes, line 606: That would be poor resolution. I suggest to use float instead. And you also can't have NaN as a default for uint.

julianoes (changes requested):

PR #182, opened 2020-10-07
#1207 extra fields for fixedwing metrics

haydockjp:
There are three extra fields in fixed wing metrics that are not returned.

- ground speed
- altitude
- heading


julianoes:
@haydockjp thanks for updating this. Now the one question remaining for me is if these fields are required or already covered by other messages/APIs.

The altitude we already have in https://github.com/mavlink/MAVSDK/blob/3e63184597eef957abc9c259a94322b5bf33d8e8/src/plugins/telemetry/include/plugins/telemetry/telemetry.h#L144-L147,

and heading/ground speed is exposed in https://github.com/mavlink/MAVSDK/blob/3e63184597eef957abc9c259a94322b5bf33d8e8/src/plugins/telemetry/include/plugins/telemetry/telemetry.h#L577-L581, you just have to do use the tangens manually to get heading, right?

PR #189, opened 2020-12-17
Adds custom_action proto

TSC21:
Adds proto to be used by the custom_action plugin. PR to MAVSDK to follow.


TSC21:
@julianoes @JonasVautherin FYI and review.

PR #190, opened 2021-01-26
Added DoOrbit action

DoppleGangster:


protos/action/action.proto
```diff
@@ -80,6 +80,12 @@ service ActionService {
      * The yaw angle is in degrees (frame is NED, 0 is North, positive is clockwise).
      */
     rpc GotoLocation(GotoLocationRequest) returns(GotoLocationResponse) {}
+    /*
+     *Send command do orbit to the drone.
```
JonasVautherin, line 84:
```suggestion
     * Send command do orbit to the drone.
```

protos/action/action.proto
```diff
@@ -80,6 +80,12 @@ service ActionService {
...
+     *
+     *This will run the do orbit routine with the given parameters.
```
JonasVautherin, line 86:
```suggestion
     * This will run the orbit routine with the given parameters.
```

protos/action/action.proto
```diff
@@ -177,6 +183,18 @@ message GotoLocationResponse {
     ActionResult action_result = 1;
 }
 
+message DoOrbitRequest {
+    float radius_m = 1; // Radius of circle (in meters)
+    float velocity_ms = 2; // Tangential velocity (in m/s)
+    float yaw_behavior = 3; // Yaw behavior of vehicle (ORBIT_YAW_BEHAVIOUR)
```
JonasVautherin, line 189: Hmm this should not be a float, but an enum, corresponding to what PX4 implements from the [MAVLink definition](http://mavlink.io/en/messages/common.html#ORBIT_YAW_BEHAVIOUR).

protos/action/action.proto
```diff
@@ -177,6 +183,18 @@ message GotoLocationResponse {
...
+    double latitude_x = 5; //Center point latitude (no mav_frame) / X coord to mav_frame. NAN: current
+    double longitude_y = 6; //Center point longitude (no mav_fram) / Y coord to mav_frame. NAN: current
+    double altitude_z = 7; //Center point altitude (MSL, no mav_frame) / Z coord to mav_frame. NAN: current
```
JonasVautherin, line 192:
```suggestion
    double latitude_x = 5; // Center point latitude (no mav_frame) / X coord to mav_frame. NAN: current
    double longitude_y = 6; // Center point longitude (no mav_fram) / Y coord to mav_frame. NAN: current
    double altitude_z = 7; // Center point altitude (MSL, no mav_frame) / Z coord to mav_frame. NAN: current
```

I'm not sure I get that. So it's either a latitude in degrees, or an x coordinate in meters w.r.t. to some local frame?

JonasVautherin (changes requested):
Thanks! Just a few comments and questions :slightly_smiling_face:

protos/action/action.proto
DoppleGangster: I was just trying to abbreviate the description as in this example:

"Center point latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. NaN: Use current vehicle position or current center if already orbiting."

protos/action/action.proto
DoppleGangster: I updated the file to what I believe is the correct way to implement the enum.

DoppleGangster:
I've commented/updated as requested. Thanks.

protos/action/action.proto
JonasVautherin: Where does that example come from?

protos/action/action.proto
```diff
@@ -177,6 +183,26 @@ message GotoLocationResponse {
...
 }
 
+message DoOrbitRequest {
+    enum OrbitYawBehavior {
+        ORBIT_YAW_BEHAVIOUR_HOLD_FRONT_TO_CIRCLE_CENTER = 0;
+        ORBIT_YAW_BEHAVIOUR_HOLD_INITIAL_HEADING = 1;
+        ORBIT_YAW_BEHAVIOUR_UNCONTROLLED = 2;
+        ORBIT_YAW_BEHAVIOUR_HOLD_FRONT_TANGENT_TO_CIRCLE = 3;
+        ORBIT_YAW_BEHAVIOUR_RC_CONTROLLED = 4;
+    }
```
JonasVautherin, line 193: Nice! However, I think the code won't generate properly with the enum inside of `DoOrbitRequest` (did you try to generate the corresponding code?). The simplest fix would be to move the enum out of it (e.g. to line 185, or at the end of the file, after the *Request definitions).

So the fix is very easy: just move this enum out of this message :blush:.

JonasVautherin (changes requested):

protos/action/action.proto
DoppleGangster: Cool, done.

DoppleGangster:
Alrighty hopefully third times the charm.

protos/action/action.proto
```diff
@@ -177,6 +183,26 @@ message GotoLocationResponse {
     ActionResult action_result = 1;
 }
 
+enum OrbitYawBehavior {
+        ORBIT_YAW_BEHAVIOUR_HOLD_FRONT_TO_CIRCLE_CENTER = 0;
+        ORBIT_YAW_BEHAVIOUR_HOLD_INITIAL_HEADING = 1;
+        ORBIT_YAW_BEHAVIOUR_UNCONTROLLED = 2;
+        ORBIT_YAW_BEHAVIOUR_HOLD_FRONT_TANGENT_TO_CIRCLE = 3;
+        ORBIT_YAW_BEHAVIOUR_RC_CONTROLLED = 4;
```
JonasVautherin, line 191:
```suggestion
    ORBIT_YAW_BEHAVIOUR_HOLD_FRONT_TO_CIRCLE_CENTER = 0;
    ORBIT_YAW_BEHAVIOUR_HOLD_INITIAL_HEADING = 1;
    ORBIT_YAW_BEHAVIOUR_UNCONTROLLED = 2;
    ORBIT_YAW_BEHAVIOUR_HOLD_FRONT_TANGENT_TO_CIRCLE = 3;
    ORBIT_YAW_BEHAVIOUR_RC_CONTROLLED = 4;
```

protos/action/action.proto
```diff
@@ -177,6 +183,26 @@ message GotoLocationResponse {
...
+}
+    
+message DoOrbitRequest {    
+    float radius_m = 1; // Radius of circle (in meters)
+    float velocity_ms = 2; // Tangential velocity (in m/s)
+    OrbitYawBehavior yaw_behavior = 3; // Yaw behavior of vehicle (ORBIT_YAW_BEHAVIOUR)
+    double latitude_x = 5; // Center point latitude (no mav_frame) / X coord to mav_frame. NAN: current
```
JonasVautherin, line 198: I am not really sure what this means, still. From where did you copy that?

Usually we have `latitude_deg`, I guess the x would be in meters (like `x_m` maybe?). The thing is that I am not sure if MAVSDK supports flying without having a GPS fix (yet). Did you try using `x` instead of `latitude`?

If it only works with a latitude, I would change that for `latitude_deg` and not expose the `xyz` ones.

What do you think?

JonasVautherin (changes requested):

protos/action/action.proto
DoppleGangster: Its from the MAVLINK messages for MAV_CMD_DO_ORBIT param descriptions. I believe is allows for specifying a location to orbit. If no mavframe is given and the value is treated as a GPS location. If nan is given the it just uses the position that would be at the center of the current radius. I'm guess since it doesn't have a parameter for mavframe that is an option for the future? I would be fine with removing all the options for position and just keeping it simple so that when the command is given it just starts the orbit based on the center point to the radius.

JonasVautherin: Intuitively, I would expect to set an orbit with a center and a radius. That's what we have here, so I would keep it. Except that if we don't know if that `xyz` thing works, I would suggest we only go with the GPS coordinates. That's what you are testing, isn't it?

So something like:

```
double latitude_deg = 5; // COMMENT
double longitude_deg = 6; // COMMENT
double absolute_altitude_m = 7; // COMMENT
```

Would that work for you?

DoppleGangster: That works for me. I did test inputting NAN as well and that worked as well. It just starts the orbit at the current position based on the set radius.

DoppleGangster:
Updated parameter names and descriptions as discussed.

protos/action/action.proto
```diff
@@ -177,6 +183,26 @@ message GotoLocationResponse {
...
+        ORBIT_YAW_BEHAVIOUR_RC_CONTROLLED = 4;
+}
+    
+message DoOrbitRequest {    
+    float radius_m = 1; // Radius of circle (in meters)
+    float velocity_ms = 2; // Tangential velocity (in m/s)
+    OrbitYawBehavior yaw_behavior = 3; // Yaw behavior of vehicle (ORBIT_YAW_BEHAVIOUR)
+    double latitude_deg = 5; // Center point latitude in degrees. NAN: current center
+    double longitude_deg = 6; // Center point longitude in degrees. NAN: current center
+    double absolute_altitude_m = 7; // Center point altitude in meters. NAN: current elevation
```
JonasVautherin, line 200: Let's default them to NaN, then :blush:, if you agree.

This requires you to add `import "mavsdk_options.proto";` at the top of `action.proto` (see e.g. [here](https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/mission/mission.proto#L5))

```suggestion
    double latitude_deg = 5 [(mavsdk.options.default_value)="NaN"; // Center point latitude in degrees. NAN: use current latitude for center
    double longitude_deg = 6 [(mavsdk.options.default_value)="NaN"; // Center point longitude in degrees. NAN: use current longitude for center
    double absolute_altitude_m = 7 [(mavsdk.options.default_value)="NaN"; // Center point altitude in meters. NAN: use current altitude for center
```

JonasVautherin (commented):
Looks good to me! Just the nitpick for the enum indentation, and I made one last suggestion about defaulting the center to NaN :slightly_smiling_face:.

Thanks a lot for your contributions!

DoppleGangster:
No problem. Thanks for working with me.

DoppleGangster:
Added:
 
`import` "mavsdk_options.proto";`

So looks like it passed the checks this time.

JonasVautherin (approved):

JonasVautherin:
Thanks a lot! Now the next step is to update the proto submodule in your `mavlink/MAVSDK` fork to point to `mavlink/MAVSDK-Proto:master`, and open a PR with your changes in `mavlink/MAVSDK`.

Does that make sense?

DoppleGangster:
It does. I'll work on getting the PR ready. Thanks!

PR #192, opened 2021-01-31
Added comment for Orbit Yaw Behavior enum.

DoppleGangster:
Side note the enum in the documentation has behavior spelled as behaviour for Orbit_Yaw_Behaviour [here](https://mavlink.io/en/messages/common.html).

### DIFF
--- a/protos/action/action.proto
+++ b/protos/action/action.proto
@@ -185,6 +185,7 @@ message GotoLocationResponse {
     ActionResult action_result = 1;
 }
 
+// Yaw behaviour during orbit flight.
 enum OrbitYawBehavior {
     ORBIT_YAW_BEHAVIOR_HOLD_FRONT_TO_CIRCLE_CENTER = 0;
     ORBIT_YAW_BEHAVIOR_HOLD_INITIAL_HEADING = 1;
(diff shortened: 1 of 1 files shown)


protos/action/action.proto
```diff
@@ -185,6 +185,7 @@ message GotoLocationResponse {
     ActionResult action_result = 1;
 }
 
+// Yaw behaviour during orbit flight.
 enum OrbitYawBehavior {
     ORBIT_YAW_BEHAVIOR_HOLD_FRONT_TO_CIRCLE_CENTER = 0;
```
JonasVautherin, line 190: It's missing the documentation next to the values, like [here](https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/camera/camera.proto#L218-L222)

JonasVautherin (changes requested):

protos/action/action.proto
DoppleGangster: It has been enhanced.

DoppleGangster:
Should be good to go now.

JonasVautherin (approved):
Thanks!

PR #193, opened 2021-02-02
Add control messages for gimbal v2

JonasVautherin:
To be consistent, the messages will be available when the gimbal implements v1, but they will always succeed (so that the client can write the same code for both versions).

Secondary mode is not yet supported, and will return a `Gimbal::Result::Unsupported` immediately.

### DIFF
--- a/protos/gimbal/gimbal.proto
+++ b/protos/gimbal/gimbal.proto
@@ -65,7 +65,8 @@ service GimbalService {
       * Subscribe to control status updates.
       *
       * This allows a component to know if it has primary, secondary or
-      * no control over the gimbal.
+      * no control over the gimbal. Also, it gives the system and component ids
+      * of the other components in control (if any).
       */
      rpc SubscribeControl(SubscribeControlRequest) returns(stream ControlResponse) {}
 }
@@ -116,7 +117,7 @@ message ReleaseControlResponse {
 
 message SubscribeControlRequest {}
 message ControlResponse {
-    ControlMode control_mode = 1; // Control mode (none, primary or secondary)
+    ControlStatus control_status = 1; // Control status
 }
 
 // Gimbal mode type.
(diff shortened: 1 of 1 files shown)


julianoes (commented):
Should we distinguish between the case where we take control when no one is in control yet and the case where someone else is in control and we need to force it on purpose?

JonasVautherin:
Honestly I don't know. The way I was picturing it is this:

I am on the ground, looking at my GCS. I have the camera/gimbal controls showing on the screen, but "locked" to show me that I am not in control. I click on the "unlock" button, and the controls get activated. I can now move the gimbal.

Later, I start a cinematic shot, and there the companion takes control, resulting in my buttons to get disabled again (automatically). At the end of the shot, the companion releases control, but my buttons are still grayed out, just like when I started.

Now during the cinematic shot, I could actually click on the "unlock" button, and my GCS would give me secondary control (because it knows that somebody else has control).

The issue is that there are tons of weird cases that can happen: what if there are two GCS with two operators that keep on acquiring primary control? But it's super difficult to solve in a generic way, in the end a specific GCS has to decide what it does based on a specific drone and other potential specific GCS implementations (e.g. "I am implementing an Android app to control a Yuneec H520 that is also being controlled by my QGC fork in parallel, and the QGC fork has priority over my app").

protos/gimbal/gimbal.proto
```diff
@@ -77,12 +103,45 @@ message SetRoiLocationResponse {
...
+    CONTROL_MODE_SECONDARY = 2; // To take secondary control over the gimbal
+}
+
+// Control status
+message ControlStatus {
+    ControlMode control_mode = 1; // Control mode (none, primary or secondary)
+    int32 sysid_primary_control = 2; // Sysid of the component that has primary control over the gimbal
+    int32 compid_primary_control = 3; // Compid of the component that has primary control over the gimbal
+    int32 sysid_secondary_control = 4; // Sysid of the component that has secondary control over the gimbal
+    int32 compid_secondary_control = 5; // Compid of the component that has secondary control over the gimbal
```
julianoes, line 142:
```suggestion
    int32 sysid_primary_control = 2; // Sysid of the component that has primary control over the gimbal (0 if no one is in control)
    int32 compid_primary_control = 3; // Compid of the component that has primary control over the gimbal (0 if no one is in control)
    int32 sysid_secondary_control = 4; // Sysid of the component that has secondary control over the gimbal (0 if no one is in control)
    int32 compid_secondary_control = 5; // Compid of the component that has secondary control over the gimbal (0 if no one is in control)
```

PR #196, opened 2021-02-08
telemetry: add timestamps to attitude

julianoes:


protos/telemetry/telemetry.proto
```diff
@@ -421,6 +421,7 @@ message Quaternion {
     float x = 2 [(mavsdk.options.default_value)="NaN"]; // Quaternion entry 1, also denoted as b
     float y = 3 [(mavsdk.options.default_value)="NaN"]; // Quaternion entry 2, also denoted as c
     float z = 4 [(mavsdk.options.default_value)="NaN"]; // Quaternion entry 3, also denoted as d
+    uint64 timestamp_us = 5;
```
JonasVautherin, line 424: It will need a comment for the docs :)

JonasVautherin (changes requested):

protos/telemetry/telemetry.proto
julianoes: Oops.

PR #197, opened 2021-02-12
Add a way to list photos from camera plugin

JonasVautherin:
Probably need some work on the wording, but you get the idea :sweat_smile:.

I am also considering having a LAST_MISSION range, that would list the images taken during the last mission (current mission if not finished).


julianoes (approved):
Looks ok.

julianoes (approved):

PR #198, opened 2021-02-17
Add tracking_server plugin

julianoes:
This adds a plugin for onboard tracking software.

### DIFF
--- a/protos/tracking_server/tracking_server.proto
+++ b/protos/tracking_server/tracking_server.proto
@@ -53,7 +53,9 @@ message TrackingRectangleCommandResponse {
 }
 
 message SubscribeTrackingOffCommandRequest {}
-message TrackingOffCommandResponse {}
+message TrackingOffCommandResponse {
+    int32 dummy = 1; // Unused
+}
 
 message RespondTrackingPointCommandRequest {
     TrackingServerResult result = 1; // The result to answer to the incoming command
--- a/protos/tracking_server/tracking_server.proto
+++ b/protos/tracking_server/tracking_server.proto
@@ -58,32 +58,34 @@ message TrackingOffCommandResponse {
 }
 
 message RespondTrackingPointCommandRequest {
-    TrackingServerResult result = 1; // The result to answer to the incoming command
+    CommandAnswer command_answer = 1; // The ack to answer to the incoming command
 }
 message RespondTrackingPointCommandResponse {
-    TrackingServerResult tracking_server_result = 1; // The result of sending the response.
+    TrackingServerResult result = 1; // The result of sending the response.
 }
 
 message RespondTrackingRectangleCommandRequest {
-    TrackingServerResult result = 1; // The result to answer to the incoming command
+    CommandAnswer command_answer = 1; // The ack to answer to the incoming command
 }
 message RespondTrackingRectangleCommandResponse {
-    TrackingServerResult tracking_server_result = 1; // The result of sending the response.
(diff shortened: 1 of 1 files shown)


protos/tracking_server/tracking_server.proto
```diff
@@ -0,0 +1,76 @@
...
+message RespondTrackingCommandRequest {
+    TrackingMode tracking_mode = 1;
+    TrackingServerResult result = 2;
+}
+message RespondTrackingCommandResponse {
+    TrackingServerResult result = 1;
+}
+
+message TrackPoint {
+    float point_x = 1;
```
JonasVautherin, line 50: You'll need documentation there :slightly_smiling_face:

protos/tracking_server/tracking_server.proto
```diff
@@ -0,0 +1,76 @@
...
+
+import "mavsdk_options.proto";
+
+option java_package = "io.mavsdk.tracking_server";
+option java_outer_classname = "TrackingServerProto";
+
+// API for an onboard image tracking software.
+service TrackingServerService {
+    // Set/update the current tracking status.
+    rpc SetTrackingStatus(SetTrackingStatusRequest) returns(SetTrackingStatusResponse) { option (mavsdk.options.async_type) = SYNC; }
```
JonasVautherin, line 13: I would `SetPointTrackingStatus` and `SetRectangleTrackingStatus` instead of having to deal with parameters dependencies ("if I set mode to 'point', then I need to use the second param, and if I set it to 'rectangle', then I need to use the third param").

But that's just a personal preference :+1:

JonasVautherin (changes requested):
It's just missing some documentation, otherwise I guess it's fine :+1:

protos/tracking_server/tracking_server.proto
julianoes: Yes, I was considering that. Would you also do 3 requests for the commands then?

protos/tracking_server/tracking_server.proto
julianoes: Oops, thanks.

protos/tracking_server/tracking_server.proto
julianoes: And what about if tracking is off?

protos/tracking_server/tracking_server.proto
```diff
@@ -0,0 +1,106 @@
...
+message SubscribeTrackingRectangleCommandRequest {}
+message TrackingRectangleCommandResponse {
+    TrackRectangle track_rectangle = 1; // The point to track if a point is to be tracked
+}
+
+message SubscribeTrackingOffCommandRequest {}
+message TrackingOffCommandResponse {}
+
+message RespondTrackingPointCommandRequest {
+    TrackingServerResult result = 1; // The result to answer to the incoming command
```
JonasVautherin, line 59: I think it could be fine, but it could also be that the generator expects it to be called `tracking_server_result` instead of `result`. But you'll see it early if it complains :see_no_evil:

JonasVautherin (approved):

protos/tracking_server/tracking_server.proto
julianoes: Oh, right, I'll change it.

julianoes: I think that would be for the response though, not the result.

JonasVautherin (dismissed):

JonasVautherin (approved):

PR #204, opened 2021-02-23
Result name needs to correspond to the result type

JonasVautherin:
That's needed at least for Java :sweat_smile:


julianoes:
Thx!

PR #207, opened 2021-03-02
Added horizontal FOV and videostream light spectrum enum

markusachtelik:
[outdated:]
This PR adds the interfaces to subscribe to VIDEO_STREAM_STATUS messages from the camera. https://mavlink.io/en/messages/common.html#VIDEO_STREAM_STATUS 

[new]
after discussions, we decided to not duplicate the redundant mavlink messages VIDEO_STREAM_STATUS and VIDEO_STREAM_INFORMATION. Instead, we extend the current structure, and update it whenever data from either of these messages arrives. 
PR in mavsdk will follow. 
@julianoes , @JonasVautherin  can you take a look?

### DIFF
--- a/protos/camera/camera.proto
+++ b/protos/camera/camera.proto
@@ -296,8 +296,8 @@ message VideoStreamSettings {
     uint32 vertical_resolution_pix = 3; // Vertical resolution (in pixels)
     uint32 bit_rate_b_s = 4; // Bit rate (in bits per second)
     uint32 rotation_deg = 5; // Video image rotation (clockwise, 0-359 degrees)
-    float horizontal_fov_deg = 7; // Horizontal fov in degrees
     string uri = 6; // Video stream URI
+    float horizontal_fov_deg = 7; // Horizontal fov in degrees
 }
 
 // Information about the video stream.
--- a/protos/camera/camera.proto
+++ b/protos/camera/camera.proto
@@ -303,21 +303,21 @@ message VideoStreamSettings {
 // Information about the video stream.
 message VideoStreamInfo {
     // Video stream status type.
-    enum Status {
-        STATUS_NOT_RUNNING = 0; // Video stream is not running
-        STATUS_IN_PROGRESS = 1; // Video stream is running
+    enum VideoStreamStatus {
+        VIDEO_STREAM_STATUS_NOT_RUNNING = 0; // Video stream is not running
+        VIDEO_STREAM_STATUS_IN_PROGRESS = 1; // Video stream is running
     }
 
     // Video stream light spectrum type
-    enum Spectrum {
-        SPECTRUM_UNKNOWN = 0; // Unknown
-        SPECTRUM_VISIBLE_LIGHT = 1; // Visible light
-        SPECTRUM_INFRARED = 2; // Infrared
+    enum VideoStreamSpectrum {
+        VIDEO_STREAM_SPECTRUM_UNKNOWN = 0; // Unknown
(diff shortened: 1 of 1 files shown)


protos/camera/camera.proto
```diff
@@ -65,6 +65,10 @@ service CameraService {
      * Subscribe to video stream info updates.
      */
     rpc SubscribeVideoStreamInfo(SubscribeVideoStreamInfoRequest) returns(stream VideoStreamInfoResponse) {}
+    /*
+     * Subscribe to video stream info updates.
```
julianoes, line 69:
```suggestion
     * Subscribe to video stream status updates.
```

protos/camera/camera.proto
```diff
@@ -311,6 +320,18 @@ message VideoStreamInfo {
     Status status = 2; // Current status of video streaming
 }
 
+// Type for video stream status. Difference to VideoStreamSettings: Status gets sent whenever there is a change, e.g. fov of the camera. Settings get queried usually once during startup.
+message VideoStreamStatus{
+    bool is_running = 1; // Video stream is running / not running
```
julianoes, line 325: This is e.g. a duplicate with `Status` (`STATUS_NOT_RUNNING` and `STATUS_IN_PROGRESS`).

protos/camera/camera.proto
```diff
@@ -311,6 +320,18 @@ message VideoStreamInfo {
...
+    bool is_thermal = 2; // Indicates whether the stream is from thermal imaging (true) or visible light (false)
+    float frame_rate_hz = 3; // Frames per second
+    uint32 horizontal_resolution_pix = 4; // Horizontal resolution (in pixels)
+    uint32 vertical_resolution_pix = 5; // Vertical resolution (in pixels)
+    uint32 bit_rate_b_s = 6; // Bit rate (in bits per second)
+    uint32 rotation_deg = 7; // Video image rotation (clockwise, 0-359 degrees)
+    uint32 horizontal_fov_deg = 8; // Horizontal field of view in degrees
```
julianoes, line 332: These two types we should change to `float` anyway, either here or in `VideoStreamSettings` or both.

julianoes (changes requested):
We now basically have the same mess as in the mavlink messages. We have the same fields twice, once as status and one as information/settings. I don't think that's a great.

If you look at `VideoStreamInfo` it already contains dynamic fields such as `Status` (running/not running), so I would probably suggest to extent that rather than duplicate it.

For the MAVSDK API we don't necessarily need to distinguish between the static and dynamic data because we don't care as much about the overhead, or the amount of bytes needed to be transmitted, so I would say one big structure will all information is sane.

protos/camera/camera.proto
```diff
@@ -296,6 +296,7 @@ message VideoStreamSettings {
     uint32 vertical_resolution_pix = 3; // Vertical resolution (in pixels)
     uint32 bit_rate_b_s = 4; // Bit rate (in bits per second)
     uint32 rotation_deg = 5; // Video image rotation (clockwise, 0-359 degrees)
+    float horizontal_fov_deg = 7; // Horizontal fov in degrees
```
julianoes, line 299: @JonasVautherin @markusachtelik is 7 above 6 on purpose here? Something to do with trimming?

markusachtelik: leftover from some experimenting, will change

protos/camera/camera.proto
markusachtelik: `horizontal_fov_deg` changed to float. keeping `rotation_deg` to avoid breaking existing code

JonasVautherin (approved):

julianoes (approved):

PR #209, opened 2021-03-02
Only create the dictionary when it does not exist

JonasVautherin:
When parsing the docs, the script is creating its own struct to keep the documentation. When it finds a new nested enum, it creates a new dictionary to save its values. Until now, we never had more than two nested enums together, so the script worked, but the second nested enum actually erases the first one and the code crashes.

FYI @markusachtelik


julianoes (approved):
Ha, good find!

PR #210, opened 2021-03-03
Bump to 1.0.1 and add deploy CI job

JonasVautherin:
Note: setup.py does not get the version from the tag, it will just push whatever is there. So we need to make sure it's correct before we tag it. An improvement could be to have setup.py take the version number as an optional argument, and have the CI pass it when building a tag.


julianoes (approved):
Very nice.

PR #217, opened 2021-03-17
Add set actuator

coderkalyan:
@julianoes Just rebased the previous actuator PR onto main, since the feature is merged in PX4 now. If this looks good I'll do the same to the C++ library PR.

### DIFF
--- a/protos/action/action.proto
+++ b/protos/action/action.proto
@@ -99,6 +99,10 @@ service ActionService {
      */
     rpc Hold(HoldRequest) returns(HoldResponse) {}
     /*
+     * Send command to set the value of an actuator.
+     */
+    rpc SetActuator(SetActuatorRequest) returns(SetActuatorResponse) {}
+    /*
      * Send command to transition the drone to fixedwing.
      *
      * The associated action will only be executed for VTOL vehicles (on other vehicle types the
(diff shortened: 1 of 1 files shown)


protos/action/action.proto
```diff
@@ -206,6 +210,14 @@ message DoOrbitResponse {
     ActionResult action_result = 1;
 }
 
+message SetActuatorRequest {
+    int32 index = 1; // Index of actuator (starting with 1)
```
julianoes, line 214: Why starting with 1? And do you know how this number relates to the PWM outputs of e.g. a Pixhawk?

coderkalyan: @julianoes what would you prefer we start with? Unfortunately its all very implementation specific, and currently the PX4 implementation maps the first 3 fields of the SET_ACTUATOR command to AUX 1,2,3 on the mixer, which corresponds here to index = 1,2,3.

julianoes: Aha, so this "only" works for the aux outputs. I was hoping it would work for any output but I guess that's not so easy with the current mixer architecture. I'm ok with this for now then. We can always change it later if it changes.

coderkalyan: @julianoes Yeah it really depends on the autopilot-side implementation, I think this proto is generic enough that it wouldn't seem out of place if something changed, but in the current state of affairs it only works for the aux pins 1, 2, and 3.

Is this good to merge then?

julianoes: Ok.

julianoes (approved):

PR #219, opened 2021-03-24
Expose camera sensor information

JonasVautherin:
For a survey app, in order to compute the overlap between images, one needs the sensor information, which is not exposed yet. 

@julianoes Or is it exposed somewhere else and I missed it?

### DIFF
--- a/protos/camera/camera.proto
+++ b/protos/camera/camera.proto
@@ -368,4 +368,9 @@ message SettingOptions {
 message Information {
     string vendor_name = 1; // Name of the camera vendor
     string model_name = 2; // Name of the camera model
+    float focal_length_mm = 3; // Focal length
+    float horizontal_sensor_size_mm = 4; // Horizontal sensor size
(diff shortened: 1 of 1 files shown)


protos/camera/camera.proto
```diff
@@ -368,4 +368,9 @@ message SettingOptions {
 message Information {
     string vendor_name = 1; // Name of the camera vendor
     string model_name = 2; // Name of the camera model
+    float focal_length_mm = 3; // Focal length
```
JonasVautherin, line 371: How does it work when the camera can zoom? Does this represent the focal_length without any zoom?

Internally, should MAVSDK get the focal length coming from the camera_definition if it exists? I think this is not a standardized parameter, so that may be difficult... Or we leave it to the user to know whether to use this value or the one from the camera parameters?

julianoes (approved):
Makes sense.

julianoes (approved):

PR #220, opened 2021-03-25
Add ServerUtility plugin

julianoes:
This should be a plugin for common tasks required when running MAVSDK onboard a companion computer.


julianoes:
Self merging, can be improved later.

PR #226, opened 2021-04-29
Added new vehicle action into MissionItem message template

rligocki:
Based on issue: https://github.com/mavlink/MAVSDK-Swift/issues/198


protos/mission/mission.proto
```diff
@@ -182,6 +183,16 @@ message MissionItem {
         CAMERA_ACTION_START_VIDEO = 4; // Start capturing video
         CAMERA_ACTION_STOP_VIDEO = 5; // Stop capturing video
     }
+
+    enum PlaneAction {
+        VEHICLE_ACTION_NONE = 0;
+        VEHICLE_ACTION_ARM = 1;
+        VEHICLE_ACTION_DISARM = 2;
```
julianoes, line 190: I don't think we should have these in here. To me these are safety critical and should be explicitly used from the action plugin.

protos/mission/mission.proto
```diff
@@ -182,6 +183,16 @@ message MissionItem {
...
+        VEHICLE_ACTION_TAKEOFF = 3;
+        VEHICLE_ACTION_LAND = 4;
```
julianoes, line 192: Are these usable by multicopter as well? If so, this should be called VehicleAction (everwhere) and not PlaneAction.

rligocki: I agree, I forgot to rename it, in latest commit ist is already named VehicleAction.

protos/mission/mission.proto
```diff
@@ -172,6 +172,7 @@ message MissionItem {
     CameraAction camera_action = 8; // Camera action to trigger at this mission item
     float loiter_time_s = 9 [(mavsdk.options.default_value)="NaN"]; // Loiter time (in seconds)
     double camera_photo_interval_s = 10 [(mavsdk.options.default_value)="1.0"]; // Camera photo interval to use after this mission item (in seconds)
+    VehicleAction vehicle_action = 11; // Plane action to trigger at this mission item. Same idea as CameraAction but for plane
```
JonasVautherin, line 175:
```suggestion
    VehicleAction vehicle_action = 11; // Vehicle action to trigger at this mission item
```

JonasVautherin:
Nitpick on the doc comment, but otherwise it looks good to me! The next step now would be to generate the C++ code and implement that new action :blush:.

Note that we will not merge this PR until the C++ implementation is done, that's normal. But now you know that we seem to agree on the interface, and you can move on to the actual implementation :+1:

protos/mission/mission.proto
```diff
@@ -182,6 +183,14 @@ message MissionItem {
         CAMERA_ACTION_START_VIDEO = 4; // Start capturing video
         CAMERA_ACTION_STOP_VIDEO = 5; // Stop capturing video
     }
+
```
JonasVautherin, line 186: Missing the documentation here (look at `enum CameraAction` above)

JonasVautherin (changes requested):

julianoes:
@rligocki sorry, it looks like we were too slow to merge this. Do you mind fixing the conflict.

rligocki:
It should be fixed. Commit [48b5349](https://github.com/mavlink/MAVSDK-Proto/pull/226/commits/48b534944ecdf77257d4f733c4185629ca6c791a) did this.

julianoes:
@rligocki ok but this PR needs updating then, I think.

JonasVautherin:
Oh but now it depends on the `acceptance_radius` PR I think :thinking:

JonasVautherin (approved):

PR #228, opened 2021-05-11
protos: add RESULT_NO_SYSTEM to all plugins

julianoes:
This adds the result for when no system is connected to all plugins that did not previously have that. This is required in mavsdk_server if no system is connected but we still need to answer something over gRPC.


JonasVautherin (approved):
Ok!

PR #234, opened 2021-06-15
Fix example protoc command

### LINKED ISSUE #233: "mavsdk_options.proto: File not found" error with default example
Running the example in http://github.com/mavlink/MAVSDK-Proto/blob/790d2c1/pb_plugins/README.rst :
```
export TEMPLATE_PATH="$(pwd)/templates/swift"
protoc --plugin=protoc-gen-custom=$(which protoc-gen-mavsdk) \
        -I../protos/action \
        --custom_out=. \
        --custom_opt=swift \
        ../protos/action/action.proto

```
results in an error:
```
mavsdk_options.proto: File not found.
action.proto: Import "mavsdk_options.proto" was not found or had errors.
```
The reason for this is the `-I` flag only included the protos/action directory, but the mavsdk_options.proto file is in the protos directory. A possible fix would be changing the include flag `-I../protos/action` to `-I../protos`

alielbashir:
fixes #233

### DIFF
--- a/pb_plugins/README.rst
+++ b/pb_plugins/README.rst
@@ -18,8 +18,8 @@ How to use it
 
     export TEMPLATE_PATH="$(pwd)/templates/swift"
     protoc --plugin=protoc-gen-custom=$(which protoc-gen-mavsdk) \
-            -I../protos/action \
+            -I../protos \
             --custom_out=. \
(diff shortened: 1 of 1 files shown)


julianoes (approved):
Makes sense.

PR #236, opened 2021-06-19
GitHub Action to lint Python code

cclauss:
Output: https://github.com/cclauss/MAVSDK-Proto/actions

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -1,0 +1,11 @@
+name: lint_python
+on: [pull_request, push]
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install codespell flake8
+      - run: codespell --ignore-words-list="ned,positionned" # --skip=""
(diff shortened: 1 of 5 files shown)


JonasVautherin:
Thanks for the suggestion! I get the idea, but I'm a bit reluctant to add a linter without committing to making it pass.

For instance, it complains about typo like "NED -> NEED", but NED means "North-East-Down". So one would have to solve that somehow. Same for all the other checks.

Happy to reconsider it if somebody puts resources on making the tests pass, though :+1:

TSC21:
> Thanks for the suggestion! I get the idea, but I'm a bit reluctant to add a linter without committing to making it pass.
> 
> For instance, it complains about typo like "NED -> NEED", but NED means "North-East-Down". So one would have to solve that somehow. Same for all the other checks.
> 
> Happy to reconsider it if somebody puts resources on making the tests pass, though 👍

You would need to add a custom dictionary to make it pass.

cclauss:
`codespell --ignore-words-list="ned,positionned"`

JonasVautherin (approved):
Very nice, thanks! Other tests can be added later :blush:

julianoes:
We have spelling mistakes, no way! :smile:

PR #238, opened 2021-07-02
Add param_server

bazfp:
API for setting/getting the parameters in MAVSDK as a parameter server.

Implementation Here: https://github.com/mavlink/MAVSDK/pull/1473

### DIFF
--- a/protos/param_server/param_server.proto
+++ b/protos/param_server/param_server.proto
@@ -7,74 +7,74 @@ import "mavsdk_options.proto";
 option java_package = "io.mavsdk.param_server";
 option java_outer_classname = "ParamServerProto";
 
-// Provide raw access to get and set parameters.
+// Provide raw access to retrieve and provide server parameters.
 service ParamServerService {
     /*
-     * Get an int parameter.
+     * Retrieve an int parameter.
      *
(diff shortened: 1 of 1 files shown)


julianoes (commented):
Thanks for the pull request. So I'm a wondering about the wording because we are not really "getting" or "setting" a param but rather "providing" a param and then getting an update on it, right? I guess I don't have a better suggestion, I was just hoping it would be somehow distinguishable from the non-server version.

bazfp:
Yes I have thought about the same thing. I think setting and getting makes sense as we're setting/getting things on the `param_server` side instead of on the vehicle side - but the API can be confusing as it's the same as the `param` plugin.

Alternative options are `provide` `retrieve`. I could perhaps even expand the API to act more CRUD-like `create` `read` `update` `delete` - Although `creating` and `deleting` is likely to be of limited use?

Open to suggestions.

julianoes:
@JonasVautherin any opinion?
I like `provide`, `retrieve`.

julianoes (approved):

julianoes:
@JonasVautherin any objections before I merge this?

JonasVautherin (approved):
Looking very clean :+1:

PR #239, opened 2021-07-13
telemetry: Added id field to Battery message.

jffortuna:
Added id field to Battery message.
In preparation for a PR in the MAVSDK repo that will enable handling of multiple batteries.

### DIFF
--- a/protos/telemetry/telemetry.proto
+++ b/protos/telemetry/telemetry.proto
@@ -538,8 +538,9 @@ enum FixType {
 
 // Battery type.
 message Battery {
-    float voltage_v = 1 [(mavsdk.options.default_value)="NaN"]; // Voltage in volts
-    float remaining_percent = 2 [(mavsdk.options.default_value)="NaN"]; // Estimated battery remaining (range: 0.0 to 1.0)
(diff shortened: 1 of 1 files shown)


protos/telemetry/telemetry.proto
```diff
@@ -538,8 +538,9 @@ enum FixType {
 
 // Battery type.
 message Battery {
-    float voltage_v = 1 [(mavsdk.options.default_value)="NaN"]; // Voltage in volts
-    float remaining_percent = 2 [(mavsdk.options.default_value)="NaN"]; // Estimated battery remaining (range: 0.0 to 1.0)
+    uint32 id = 1 [(mavsdk.options.default_value)="0"]; // Battery ID, for systems with multiple batteries
```
JonasVautherin, line 541: Let's keep the other numbers and call this `3` instead of `1`. So that it won't break compatibility as far as gRPC is concerned.

You can keep the id where it is (first line), but just use `= 3` and leave the others unchanged :+1:). That won't change anything in the c++ code, it's really just a protobuf thing.

JonasVautherin (changes requested):

JonasVautherin (approved):

PR #240, opened 2021-07-14
Add telemetry_server proto

bazfp:
Add telemetry server proto.

This is based on the existing telemetry proto subscriptions - with some removals.

### DIFF
--- a/protos/telemetry_server/telemetry_server.proto
+++ b/protos/telemetry_server/telemetry_server.proto
@@ -1,0 +1,464 @@
+syntax = "proto3";
+
+package mavsdk.rpc.telemetry_server;
+
+import "mavsdk_options.proto";
+
+option java_package = "io.mavsdk.telemetry_server";
+option java_outer_classname = "TelemetryServerProto";
+
+/*
+ * Allow users to provide vehicle telemetry and state information
+ * (e.g. battery, GPS, RC connection, flight mode etc.) and set telemetry update rates.
+ */
+service TelemetryServerService {
+    // Publish to 'position' updates.
+    rpc PublishPosition(PublishPositionRequest) returns(PublishTelemetryResponse) {}
+    // Publish to 'home position' updates.
+    rpc PublishHome(PublishHomeRequest) returns(PublishTelemetryResponse) {}
+    // Publish to armed updates.
+    rpc PublishArmed(PublishArmedRequest) returns(PublishTelemetryResponse) {}
+    // Publish to 'Raw GPS' updates.
+    rpc PublishRawGps(PublishRawGpsRequest) returns(PublishTelemetryResponse) {}
+    // Publish to 'battery' updates.
+    rpc PublishBattery(PublishBatteryRequest) returns(PublishTelemetryResponse) {}
+    // Publish to 'flight mode' updates.
+    rpc PublishFlightMode(PublishFlightModeRequest) returns(PublishTelemetryResponse) {}
+    // Publish to 'health' updates.
+    rpc PublishHealth(PublishHealthRequest) returns(PublishTelemetryResponse) {}
+    // Publish to 'status text' updates.
+    rpc PublishStatusText(PublishStatusTextRequest) returns(PublishTelemetryResponse) {}
+    // Publish to 'odometry' updates.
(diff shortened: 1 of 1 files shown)


protos/telemetry_server/telemetry_server.proto
```diff
@@ -0,0 +1,464 @@
...
+    // Publish to 'Raw GPS' updates.
+    rpc PublishRawGps(PublishRawGpsRequest) returns(PublishTelemetryResponse) {}
+    // Publish to 'battery' updates.
+    rpc PublishBattery(PublishBatteryRequest) returns(PublishTelemetryResponse) {}
+    // Publish to 'flight mode' updates.
+    rpc PublishFlightMode(PublishFlightModeRequest) returns(PublishTelemetryResponse) {}
+    // Publish to 'health' updates.
+    rpc PublishHealth(PublishHealthRequest) returns(PublishTelemetryResponse) {}
+    // Publish to 'status text' updates.
+    rpc PublishStatusText(PublishStatusTextRequest) returns(PublishTelemetryResponse) {}
```
JonasVautherin, line 30: Hmm I think that this won't work with the autogen. `PublishStatusText` should have a `PublishStatusTextRequest` and a `PublishStatusTextResponse`. Same for all the others :+1:

bazfp: What will fail? It's compiling and running at the moment

JonasVautherin: Yes, but I know it's a convention I took in the autogen plugin `protoc-gen-mavsdk`. Here it compiles maybe because the response is empty, or maybe because C++ does not require that, but some language bindings will probably need it.

`protoc-gen-mavsdk` leverages `protoc`, and therefore there is no way to change how the proto files look. The convention I took was that a function `functionName()` is defined as `rpc FunctionName(FunctionNameRequest) returns(FunctionNameResponse) {}`.

Maybe it could be done without it, but we would need to try all the language bindings before concluding it :see_no_evil:

protos/telemetry_server/telemetry_server.proto
```diff
@@ -131,6 +129,73 @@ message PublishUnixEpochTimeRequest {
     uint64 time_us = 1; // The next 'unix epoch time' status
 }
 
+// Publish Responses
```
JonasVautherin, line 132:
```suggestion
```

Request and Response messages don't need documentation (it's not parsed, the rest of the documentation is actually used by the generated code). See [here](https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/telemetry/telemetry.proto#L150-L153) for instance, there is no documentation around it :+1:

protos/telemetry_server/telemetry_server.proto
```diff
@@ -131,6 +129,73 @@ message PublishUnixEpochTimeRequest {
...
+message PublishHealthAllOkResponse {
+    TelemetryServerResult telemetry_server_result = 1;
+}
+
+message PublishUnixEpochTimeResponse {
+    TelemetryServerResult telemetry_server_result = 1;
+}
+
+
+// GPS Origin
```
JonasVautherin, line 198:
```suggestion
```

Same here

protos/telemetry_server/telemetry_server.proto
```diff
@@ -0,0 +1,529 @@
...
+    rpc PublishHealthAllOk(PublishHealthAllOkRequest) returns(PublishHealthAllOkResponse) {}
+    // Publish to 'unix epoch time' updates.
+    rpc PublishUnixEpochTime(PublishUnixEpochTimeRequest) returns(PublishUnixEpochTimeResponse) {}
+
+    // Get the GPS location of where the estimator has been initialized.
+    rpc GetGpsGlobalOrigin(GetGpsGlobalOriginRequest) returns(GetGpsGlobalOriginResponse) {}
+}
+
+
+// Request Messages
```
JonasVautherin, line 53:
```suggestion
```

JonasVautherin (changes requested):
Some nitpicks, but otherwise it looks good to me! Let's see with the C++ code then!

julianoes (approved):

PR #242, opened 2021-07-23
Add Mission server

bazfp:
Add types required for mission_server.

### DIFF
--- a/protos/mission_server/mission_server.proto
+++ b/protos/mission_server/mission_server.proto
@@ -1,0 +1,87 @@
+syntax = "proto3";
+
+package mavsdk.rpc.mission_server;
+
+import "mavsdk_options.proto";
+
+option java_package = "io.mavsdk.mission_server";
+option java_outer_classname = "MissionServerProto";
+
+// Enable waypoint missions.
+service MissionService {
+    /*
+     * Subscribe to when a new mission is uploaded (asynchronous).
+     */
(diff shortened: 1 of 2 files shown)


protos/mission_server/mission_server.proto
```diff
@@ -0,0 +1,100 @@
+syntax = "proto3";
+
+package mavsdk.rpc.mission_server;
+
+import "mavsdk_options.proto";
+
+option java_package = "io.mavsdk.mission_server";
+option java_outer_classname = "MissionServerProto";
+
+// Enable waypoint missions.
```
julianoes, line 10: Please describe that this is a server so a vehicle side.

protos/mission_server/mission_server.proto
```diff
@@ -0,0 +1,100 @@
...
+message SubscribeClearAllRequest {}
+message ClearAllResponse {
+    uint32 clear_type = 1;
+}
+
+message SetCurrentItemCompleteRequest {}
+message SetCurrentItemCompleteResponse {}
+
+// Mission item exactly identical to MAVLink MISSION_ITEM_INT.
+message MissionItem {
```
julianoes, line 53: I suppose this should be a `MissionItemRaw` to keep the symmetry with the other plugins? Maybe even the plugin should be called `mission_raw_server`.

bazfp: Mission Raw plugin has the same type naming scheme here. I will change the plugin to `mission_raw_server` for parity.

julianoes (approved):

PR #250, opened 2021-09-01
added vtol state to telemetry (follow up)

mattes-bru:
Added information of the current vtol state to the telemetry plugin

### DIFF
--- a/protos/telemetry/telemetry.proto
+++ b/protos/telemetry/telemetry.proto
@@ -22,6 +22,8 @@ service TelemetryService {
     rpc SubscribeLandedState(SubscribeLandedStateRequest) returns(stream LandedStateResponse) {}
     // Subscribe to armed updates.
     rpc SubscribeArmed(SubscribeArmedRequest) returns(stream ArmedResponse) {}
+    // subscribe to vtol state Updates
+    rpc SubscribeVtolState(SubscribeVtolStateRequest) returns(stream VtolStateResponse) {}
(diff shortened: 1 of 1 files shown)


julianoes:
Hang on, we just added that like yesterday :joy: https://github.com/mavlink/MAVSDK-Proto/pull/249, oops.

protos/telemetry/telemetry.proto
```diff
@@ -85,6 +87,8 @@ service TelemetryService {
     rpc SetRateInAir(SetRateInAirRequest) returns(SetRateInAirResponse) {}
     // Set rate to landed state updates
     rpc SetRateLandedState(SetRateLandedStateRequest) returns(SetRateLandedStateResponse) {}
+    // Set rate to VTOL state updates
+    rpc SetRateVtolState(SetRateVtolStateRequest) returns(SetRateVtolStateResponse) {}
```
julianoes, line 91: I would actually not expose that because it's tied to the status anyway, and generally I'd say 1 Hz is good enough.

mattes-bru: I didn't think much about it, I did more or less a simply copy of the other information extracted from the `EXTENDED_SYS_STATE`

julianoes: Yer makes sense.

julianoes:
I've removed the conflicts and force-pushed it.

PR #254, opened 2021-09-05
Fix protos for java

JonasVautherin:
This should have no impact on C++, but the inconsistencies in the proto files were breaking the java build. Those things could be enforced by protoc-gen-mavsdk, but I would need to spend some time on it.


julianoes:
Oops, thanks.

PR #255, opened 2021-09-22
Add prepare() to camera plugin

JonasVautherin:
Plugins are lazily initialized in mavsdk_server. This allows to manually initialize the camera plugin (and have it download the camera definition file, etc).


julianoes (commented):
You don't want to return a result to signal that it is now prepared? Or a status `is_prepared()` to check if it's ready?

PR #256, opened 2021-10-20
core: add API to set sysid/compid

julianoes:


JonasVautherin (commented):
Done this way, will it not start with the default `sysid:compid`, possibly talk to other components/systems over the network, and then change the `sysid:compid` when this function is called? :thinking:

PR #265, opened 2021-11-18
proto for offboard global pos altitude type

romain-chiap:
I added an altitude type for the offboard global position

### DIFF
--- a/protos/offboard/offboard.proto
+++ b/protos/offboard/offboard.proto
@@ -212,9 +212,9 @@ message PositionNedYaw {
 message PositionGlobalYaw {
     // Possible altitude options
     enum AltitudeType {
-        ALTITUDE_REL_HOME = 0; // Altitude realtive to the Home position
-        ALTITUDE_AMSL = 1; // Altitude above mean sea level (AMSL)
-        ALTITUDE_AGL = 2; // Altitude above ground level (AGL)
+        ALTITUDE_TYPE_REL_HOME = 0; // Altitude relative to the Home position
+        ALTITUDE_TYPE_AMSL = 1; // Altitude above mean sea level (AMSL)
+        ALTITUDE_TYPE_AGL = 2; // Altitude above ground level (AGL)
     }    
     double lat_deg = 1; // Latitude (in degrees)
     double lon_deg = 2; // Longitude (in degrees)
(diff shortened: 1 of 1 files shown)


protos/offboard/offboard.proto
```diff
@@ -210,10 +210,17 @@ message PositionNedYaw {
 
 // Type for position commands in Global (Latitude, Longitude, Altitude) coordinates and yaw.
 message PositionGlobalYaw {
+    // Possible altitude options
+    enum AltitudeType {
+        ALTITUDE_REL_HOME = 0; // Altitude realtive to the Home position
+        ALTITUDE_AMSL = 1; // Altitude above mean sea level (AMSL)
+        ALTITUDE_AGL = 2; // Altitude above ground level (AGL)
```
JonasVautherin, line 217:
```suggestion
        ALTITUDE_TYPE_REL_HOME = 0; // Altitude realtive to the Home position
        ALTITUDE_TYPE_AMSL = 1; // Altitude above mean sea level (AMSL)
        ALTITUDE_TYPE_AGL = 2; // Altitude above ground level (AGL)
```

JonasVautherin (changes requested):
Prefix needs to be the name of the enum :innocent:

romain-chiap:
little question for my own knowledge, do I need to run [step 1](https://mavsdk.mavlink.io/main/en/cpp/contributing/plugins.html#generate-h-and-cpp-files) every time I modify the proto?
`cmake -DBUILD_MAVSDK_SERVER=ON -Bbuild/default -H.`

Or can I go straight to `tools/generate_from_protos.sh` ?

What If the command I did was to build the plugins? Do I need to rerun step 1?
This step is very long on my laptop..

JonasVautherin:
No, step one only needs to be run once, unless you go change what's in https://github.com/mavlink/MAVSDK/tree/main/third_party, which you probably won't ever do. So most of the time you can go straight to `tools/generate_from_protos.sh` indeed :+1:.

The configure step (step 1) is very long because by default it runs the superbuild, i.e. it builds all the dependencies from source. You could also decide to install them somewhere out of the project (say `~/mavsdk_deps/install`), and then have `-DSUPERBUILD=OFF`. But then you would have to tell cmake where it could fine the dependencies (with something like `-DCMAKE_PREFIX_PATH=~/mavsdk_deps/install`). But probably you won't need that :blush:.

Also if you use MAVSDK in C++ only, you don't need to build mavsdk_server (`-DBUILD_MAVSDK_SERVER=OFF` will be much faster), and you can rely on the CI to test it for you (no need to edit the mavsdk_server tests actually, they come from when it was not completely auto-generated, but now they are not very important anymore).

JonasVautherin (approved):

romain-chiap:
Awesome thanks, that will accelerate my development time :)

PR #268, opened 2021-12-22
mission: add upload/download progress

julianoes:
This is a first try to add this to the API. I'm not convinced though, I feel like it would be much easier to just have an additional transfer subscription which is independent of the existing transfer.

I see two challenges:
1. A subscription with an argument in the request is probably not implemented, especially not with a repeated field.
2. The type of a DownloadMissionResponse is quite weird as it has a result, potentially the downloaded mission, as well as a transfer progress field. This doesn't seem quite right.

### DIFF
--- a/protos/mission/mission.proto
+++ b/protos/mission/mission.proto
@@ -16,17 +16,42 @@ service MissionService {
      * executed even if the connection is lost.
      */
     rpc UploadMission(UploadMissionRequest) returns(UploadMissionResponse) {}
+
+    /*
+     * Upload a list of mission items to the system with progress updates.
+     *
+     * The mission items are uploaded to a drone. Once uploaded the mission can be started and
+     * executed even if the connection is lost.
+     */
+    rpc SubscribeUploadMission(SubscribeUploadMissionRequest) returns(stream UploadMissionResponse) {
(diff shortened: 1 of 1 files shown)


JonasVautherin:
Hmm I'm not sure this will work. The stream implementation we have supports 3 kinds of results: SUCCESS, NEXT, and everything else is considered an error. SUCCESS means that the stream ended successfully, and NEXT means that this contains a progress update (or whatever the next element is).

Don't we want something like the calibration here? E.g. https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/calibration/calibration.proto#L65-L69?

julianoes:
Yes, my idea was that it would be the same as calibration.

protos/mission/mission.proto
```diff
@@ -100,8 +130,11 @@ message DownloadMissionRequest {}
 message DownloadMissionResponse {
     MissionResult mission_result = 1;
     MissionPlan mission_plan = 2; // The mission plan
+    TransferProgress = 3; // The transfer progress
```
JonasVautherin, line 133:
```suggestion
    TransferProgress transfer_progress = 3; // The transfer progress
```

julianoes:
Replaced by #272.

PR #269, opened 2021-12-30
mission: add subscriptions for transfer progress

julianoes:
This adds two separate subscriptions for the mission upload and download transfer progress, as a float from 0.0 to 1.0.

This would be an alternative to #268.


julianoes:
Replaced by #272.

PR #272, opened 2022-01-11
Add transfer progress

JonasVautherin:


protos/mission/mission.proto
```diff
@@ -221,3 +249,17 @@ message MissionResult {
     Result result = 1; // Result enum value
     string result_str = 2; // Human-readable English string describing the result
 }
+
+// Progress data coming from mission upload, in percent.
+message ProgressData {
+    float progress = 1 [(mavsdk.options.default_value)="NaN"]; // Progress (percentage)
```
julianoes, line 255: Percentage would mean 0..100 but I'd probably just do 0..1.0, right?

protos/mission/mission.proto
```diff
@@ -221,3 +249,17 @@ message MissionResult {
...
+}
+
+// Progress data coming from mission download, in percent, or the mission itself (if the transfer succeeds).
+message ProgressDataOrMission {
+    bool has_progress = 1 [(mavsdk.options.default_value)="false"]; // Whether this ProgressData contains a 'progress' status or not
+    float progress = 2 [(mavsdk.options.default_value)="NaN"]; // Progress (percentage)
+
+    bool has_mission_plan = 3; // Whether this ProgressData contains a 'mission_plan' or not
+    MissionPlan mission_plan = 4; // Mission plan
+}
```
julianoes, line 265: I'm confused, this is not used anywhere...

protos/mission/mission.proto
JonasVautherin: This is the same as the calibration, but I really don't mind. I'll change it to `Progress (0..1.0)` :+1:

julianoes (dismissed):
Yes, now it makes more sense!

julianoes (approved):

PR #273, opened 2022-01-29
protoc_gen_mavsdk: error on missing return value

dayjaby:
When defining a rpc function with a return stream, it is necessary that
this function returns something that is not a *Result. So the Response
proto must contain more than just the result entry. Here we give an
explicit error message, so that the developer knows precisely what is
missing.


JonasVautherin (approved):
Nice, thanks a lot!

PR #274, opened 2022-02-14
add new camera server

dlech:
This adds new camera server plugin protocol with a few basic commands for device discovery and still image capture.

It works like this:

1. New camera service instance is created and attached to system.
2. The Subscribe* methods are called to register callbacks. These will define the camera capability flags. If a callback is registered, then the capability flag is set, otherwise the flag is not set.
3. The Set* methods are called to set the initial state.
2. The SetInformation method is called to "activate" the plugin. This method must be called after all of the above. Also, all of the above steps must be done right away after creating the mavsdk instance, otherwise systems might think that there is no camera.
3. To handle taking pictures, the SubscribeTakePhoto method is called. When this method is called back, the implementer must call the SetInProgress with `true` method right away. Then the picture is taken and the implementer calls SetInProgress with `false`. Finally,  the implementer must call the RespondTakePhoto method.

### DIFF
--- a/protos/camera_server/camera_server.proto
+++ b/protos/camera_server/camera_server.proto
@@ -56,11 +56,15 @@ message RespondTakePhotoResponse {
 message Information {
     string vendor_name = 1; // Name of the camera vendor
     string model_name = 2; // Name of the camera model
-    float focal_length_mm = 3; // Focal length
-    float horizontal_sensor_size_mm = 4; // Horizontal sensor size
-    float vertical_sensor_size_mm = 5; // Vertical sensor size
-    uint32 horizontal_resolution_px = 6; // Horizontal image resolution in pixels
-    uint32 vertical_resolution_px = 7; // Vertical image resolution in pixels
+    string firmware_version = 3; // Camera firmware version in <major>[.<minor>[.<patch>[.<dev>]]] format
+    float focal_length_mm = 4; // Focal length
+    float horizontal_sensor_size_mm = 5; // Horizontal sensor size
+    float vertical_sensor_size_mm = 6; // Vertical sensor size
+    uint32 horizontal_resolution_px = 7; // Horizontal image resolution in pixels
+    uint32 vertical_resolution_px = 8; // Vertical image resolution in pixels
+    uint32 lens_id = 9; // Lens ID
+    uint32 definition_file_version = 10; // Camera definition file version (iteration)
+    string definition_file_uri = 11; // Camera definition URI (http or mavlink ftp)
 }
 
 // Position type in global coordinates.
--- a/protos/camera_server/camera_server.proto
+++ b/protos/camera_server/camera_server.proto
@@ -1,0 +1,118 @@
+syntax = "proto3";
+
+package mavsdk.rpc.camera_server;
+
+import "mavsdk_options.proto";
+
+option java_package = "io.mavsdk.camera_server";
+option java_outer_classname = "CameraServerProto";
+
+// Provides handling of camera trigger commands.
+service CameraServerService {
+    // Sets the camera information
+    rpc SetInformation(SetInformationRequest) returns(SetInformationResponse) { option (mavsdk.options.async_type) = SYNC; }
+    // Sets the camera capture status
+    rpc SetInProgress(SetInProgressRequest) returns(SetInProgressResponse) { option (mavsdk.options.async_type) = SYNC; }
+
+    // Subscribe to single-image capture commands
+    rpc SubscribeTakePhoto(SubscribeTakePhotoRequest) returns(stream TakePhotoResponse) { option (mavsdk.options.async_type) = ASYNC; }
+    
+    // Adds a photo to the list of available photos
+    rpc PublishPhoto(PublishPhotoRequest) returns(PublishPhotoResponse) { option (mavsdk.options.async_type) = SYNC; }
+}
+
+message SetInformationRequest {
+    Information information = 1; // information about the camera
+}
+
+message SetInformationResponse {
+    CameraServerResult camera_server_result = 1;
+}
+
+message SetInProgressRequest {
+    bool in_progress = 1; // true if capture is in progress or false for idle.
+}
+
+message SetInProgressResponse {
+    CameraServerResult camera_server_result = 1;
+}
+
+message SubscribeTakePhotoRequest {}
+
+message TakePhotoResponse {
+    CameraServerResult camera_server_result = 1;
+    int32 index = 2;
+}
+
+message PublishPhotoRequest {
+    CaptureInfo capture_info = 1;
+}
+
+message PublishPhotoResponse {
+    CameraServerResult camera_server_result = 1;
+}
+
+// Type to represent a camera information.
+message Information {
+    string vendor_name = 1; // Name of the camera vendor
+    string model_name = 2; // Name of the camera model
+    float focal_length_mm = 3; // Focal length
+    float horizontal_sensor_size_mm = 4; // Horizontal sensor size
+    float vertical_sensor_size_mm = 5; // Vertical sensor size
+    uint32 horizontal_resolution_px = 6; // Horizontal image resolution in pixels
+    uint32 vertical_resolution_px = 7; // Vertical image resolution in pixels
+}
+
+// Position type in global coordinates.
+message Position {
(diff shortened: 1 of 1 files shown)


dlech:
This is a revision of #262 based on the feedback there. Cc: @dayjaby.

protos/camera_server/camera_server.proto
```diff
@@ -0,0 +1,118 @@
...
+// Type to represent a camera information.
+message Information {
+    string vendor_name = 1; // Name of the camera vendor
+    string model_name = 2; // Name of the camera model
+    float focal_length_mm = 3; // Focal length
+    float horizontal_sensor_size_mm = 4; // Horizontal sensor size
+    float vertical_sensor_size_mm = 5; // Vertical sensor size
+    uint32 horizontal_resolution_px = 6; // Horizontal image resolution in pixels
+    uint32 vertical_resolution_px = 7; // Vertical image resolution in pixels
+}
```
dlech, line 64: One thing @dayjaby mentioned already in https://github.com/dlech/MAVSDK/pull/1 is that this list is incomplete. I don't remember why I did it this way, but a complete list would look like:

```proto
message Information {
    uint32 firmware_version = 3; // Version of the camera firmware
    float sensor_size_h_mm = 5; // Horizontal sensor size
    float sensor_size_v_mm = 6; // Vertical sensor size
    uint32 resolution_h_px = 7; // Horizontal image resolution in pixels
    uint32 resolution_v_px = 8; // Vertical image resolution in pixels
    uint32 cam_definition_version = 10; // Camera definition version (iteration)
    string cam_definition_uri = 11; // Camera definition URI (http or mavlink ftp)
}
```

dlech:
The bigger design pattern that can be inferred here is that for any state, there is a `SetXyz` method where the user code pushes the current state to the plugin. The plugin stores this state and will use it to automatically respond to request messages without calling back to user code. So future methods would be `SetCameraSettings`, `SetVideoStreamInformation`, `SetVideoStreamStatus`, `SetStorageInformation`, etc. Some of these can be made more granular to only change certain aspects of a setting instead of corresponding directly to a mavlink message. For example, the `SetInProgress` method only updates the `image_status` field of the [CAMERA_CAPTURE_STATUS](https://mavlink.io/en/messages/common.html#CAMERA_CAPTURE_STATUS) message.

For messages received from an autopilot/GCS that require the camera to actually do something and not just return the current state, there are `SubscribeXyz` methods. These are things like taking a picture, changing the mode, formatting the storage, etc. For each `SubscirbeXyz` method, there will be a corresponding `RespondXyz` method similar to how the `tracking_server` plugin works. The case of taking pictures is a bit different since there isn't a single response, so it has a `PublishPhoto` method instead of a `RespondTakePhoto` method. However, now that I am writing about it and thinking about the implementation, this may not actually be a special case after all, in which case I would rename the method to match the proposed pattern.

protos/camera_server/camera_server.proto
dlech: Is adding new fields like this in the future considered a breaking change? Or will existing code just set the new fields to the default value?

dlech:
> However, now that I am writing about it and thinking about the implementation, this may not actually be a special case after all, in which case I would rename the method to match the proposed pattern.

I've updated the proposed protocol to reflect this. The idea is that from the point of view of the plugin implementer, the subscribe function will always look like a single image capture message. The plugin internals will automatically call back the subscribe method multiple times for a single mavlink message with a time interval.

protos/camera_server/camera_server.proto
dayjaby: As far as I remember from protobuf, is that it has excellent downwards compatibility as long as you only **add** new fields and don't remove or re-order already existing fields.

dayjaby:
Some additions on top of your .proto that I had in mind can be found at https://github.com/dayjaby/MAVSDK-Proto/commits/camera-server:

1. Add image/video/survey camera modes:

```
    // Subscribe to camera mode commands
    rpc SubscribeSetCameraModeImage(SubscribeSetCameraModeImageRequest) returns(stream SetCameraModeImageResponse) {
        option (mavsdk.options.async_type) = ASYNC;
    }
    rpc SubscribeSetCameraModeVideo(SubscribeSetCameraModeVideoRequest) returns(stream SetCameraModeVideoResponse) {
        option (mavsdk.options.async_type) = ASYNC;
    }
    rpc SubscribeSetCameraModeSurvey(SubscribeSetCameraModeSurveyRequest) returns(stream SetCameraModeSurveyResponse) {
        option (mavsdk.options.async_type) = ASYNC;
    }
```

Each mode has its own subscribe function as the camera capabilities get automatically constructed depending on which camera mode subscriptions exist.

2. Zoom and focus support

```
    // Subscribe to camera zoom commands
    rpc SubscribeSetCameraZoom(SubscribeSetCameraZoomRequest) returns(stream SetCameraZoomResponse) {
        option (mavsdk.options.async_type) = ASYNC;
    }

    // Subscribe to camera focus commands
    rpc SubscribeSetCameraFocus(SubscribeSetCameraFocusRequest) returns(stream SetCameraFocusResponse) {
        option (mavsdk.options.async_type) = ASYNC;
    }
```

The ZoomRequest contains the minimal and maximal focal length, so it's able to handle both SET_CAMERA_ZOOM with the zoom values being in focal length *or* as a percentage.

3. Handling camera parameters (float/int):

```
    rpc SubscribeParamInt(ProvideParamIntRequest) returns(stream ProvideParamIntResponse) { option (mavsdk.options.async_type) = ASYNC; }
    rpc SubscribeParamFloat(ProvideParamFloatRequest) returns(stream ProvideParamFloatResponse) { option (mavsdk.options.async_type) = ASYNC; }
```

Handles stuff like CAM_ISO, CAM_EV, or any other parameter specified in the camera definition xml.

dlech:
The subscription to feature flag idea is interesting. It took me a while to understand what you meant but now think I like it. 👍  It means that we would need an additional rule that all subscriptions need to be made before calling SetInformation. I like that better than having to specify the flags and subscribe separately.

Also, as I described above, each Subscribe method needs a corresponding Respond method so that the user code can send an appropriate response when it has handled the action. It isn't safe for the plugin to assume that all of these operations were successful and respond automatically. Also it could take some time, e.g. to actually change the mode, and the response should be delayed until after the mode has actually physically been changed in the camera.

Additionally, we will need methods like `SetCameraZoom` to tell the plugin what the current zoom level is so that it can automatically respond to, for example, `MAV_CMD_REQUEST_MESSAGE`. It could be possible that the actual zoom doesn't match the requested zoom. And we also need a way to know the initial zoom state.

protos/camera_server/camera_server.proto
dlech: I think I remember why I left stuff out now. It looks like I only included the fields that are in the Information type of `camera.proto`. This made me assume that the other fields aren't really used in practice also the fact that the mavlink spec called the lens id "reserved").

Since the camera definition file is optional, I'm wondering if it would make sense to have a separate method for specifying that.

dlech:
I think there should only be one SubscribeSetMode method since this will correspond to the [CAMERA_CAP_FLAGS_HAS_MODES](https://mavlink.io/en/messages/common.html#CAMERA_CAP_FLAGS_HAS_MODES) flag. The [CAMERA_CAP_FLAGS_CAPTURE_IMAGE](https://mavlink.io/en/messages/common.html#CAMERA_CAP_FLAGS_CAPTURE_IMAGE) flag is determined by `SubcribeTakePhoto`, etc.

protos/camera_server/camera_server.proto
dayjaby: I guess it makes sense to encourage people to use and provide the camera definition file. When this plugin is ready, it's easier than ever to provide all the functionalities necessary for a camera definition file (e.g. reacting on parameter changes)

julianoes:
@dlech how should we proceed here? There seem to be some good discussion but I'm not sure if they have been applied or if you plan on changing this PR? I'm sorry I was busy for a while and only just returning to all these PRs.

dlech:
> how should we proceed here?

I think I would like to solve https://github.com/mavlink/MAVSDK/issues/1700 first. It would make creating and consuming an API much easier.

julianoes:
Oops, I guess I merged your commits as part of #283, closing this. Let's follow up in new PRs.

dlech:
Ha, OK.

PR #282, opened 2022-04-07
mission_raw: Add support for geofence and rally points

julianoes:
Taken from here: https://github.com/mavlink/MAVSDK-Proto/compare/pr-add-rally-geofence...volansi:pr-add-rally-geofence


JonasVautherin (dismissed):

julianoes:
Gonna leave it in this plugin, for the sake of getting it in. It is still the raw mission API, so I think this is ok.

PR #287, opened 2022-05-09
Update follow_me.proto to match PX4 Upstream

junwoo091400:
## Context
Due to the ongoing refactor of Follow Me in PX4, the enumerations as well as parameters have been changed. Therefore the proto file needs to be updated as well.

Upstream PR : https://github.com/PX4/PX4-Autopilot/pull/19260

## Notes

### DIFF
--- a/protos/follow_me/follow_me.proto
+++ b/protos/follow_me/follow_me.proto
@@ -30,19 +30,21 @@ service FollowMeService {
 
 // Configuration type.
 message Config {
-    // Direction relative to the target that the vehicle should follow
-    enum FollowDirection {
-        FOLLOW_DIRECTION_NONE = 0; // Do not follow
-        FOLLOW_DIRECTION_BEHIND = 1; // Follow from behind
-        FOLLOW_DIRECTION_FRONT = 2; // Follow from front
-        FOLLOW_DIRECTION_FRONT_RIGHT = 3; // Follow from front right
-        FOLLOW_DIRECTION_FRONT_LEFT = 4; // Follow from front left
+
+    // Altitude mode to configure which altitude the follow me will assume the target to be at.
+    enum FollowAltitudeMode {
+        FOLLOW_ALTITUDE_MODE_CONSTANT = 0; // Target assumed to be mobing at a constant altitude of home position (where the vehicle armed)
+        FOLLOW_ALTITUDE_MODE_TERRAIN = 1; // Target assumed to be at the terrain level sensed by the distance sensor
+        FOLLOW_ALTITUDE_MODE_TARGET_GPS = 2; // Target GPS altitude taken into account to do 3D tracking
     }
 
-    float min_height_m = 1 [(mavsdk.options.default_value)="8.0"]; // Minimum height for the vehicle in meters (recommended minimum 8 meters)
-    float follow_distance_m = 2 [(mavsdk.options.default_value)="8.0"]; // Distance from target for vehicle to follow in meters (recommended minimum 1 meter)
-    FollowDirection follow_direction = 3; // Direction to follow in
-    float responsiveness = 4 [(mavsdk.options.default_value)="0.5"]; // How responsive the vehicle is to the motion of the target (range 0.0 to 1.0)
+    float follow_height_m = 1 [(mavsdk.options.default_value)="8.0f"]; // [m] Follow height in meters (recommended minimum 8 meters)
+    float follow_distance_m = 2 [(mavsdk.options.default_value)="8.0f"]; // [m] Follow distance to target in meters (recommended minimum 4 meter)
+    // FollowDirection follow_direction = 3; // [!DEPRECIATED!] Direction to follow target at. Exists as a legacy to old PX4 Follow-Me, but now only used for setting the follow_angle_deg. This value never gets sent to the PX4 side!
+    float responsiveness = 4 [(mavsdk.options.default_value)="0.1f"]; // How responsive the vehicle is to the motion of the target, Lower value = More responsive (range 0.0 to 1.0)
+    FollowAltitudeMode altitude_mode = 5; // Follow Altitude control mode
+    float max_tangential_vel_m_s = 6 [(mavsdk.options.default_value)="8.0f"]; // [m/s] Maximum orbit tangential velocity relative to the target, in meters per second. Higher value = More aggressive follow angle tracking.
+    float follow_angle_deg = 7 [(mavsdk.options.default_value)="180.0f"]; // [deg] Follow Angle relative to the target. 0 equals following in front of the target's direction. Angle increases in Clockwise direction, so following from right would be 90 degrees, from the left is -90 degrees, and so on.
 }
 
 // Target location for the vehicle to follow


protos/follow_me/follow_me.proto
```diff
@@ -30,19 +30,23 @@ service FollowMeService {
...
+        FOLLOW_PERSPECTIVE_BEHINDLEFT = 7 ; // Follow from behind left
+        FOLLOW_PERSPECTIVE_INVALID = 8;     // Follow perspective invalid
     }
 
-    float min_height_m = 1 [(mavsdk.options.default_value)="8.0"]; // Minimum height for the vehicle in meters (recommended minimum 8 meters)
+    float follow_height_m = 1 [(mavsdk.options.default_value)="8.0"]; // Minimum height for the vehicle in meters (recommended minimum 8 meters)
     float follow_distance_m = 2 [(mavsdk.options.default_value)="8.0"]; // Distance from target for vehicle to follow in meters (recommended minimum 1 meter)
-    FollowDirection follow_direction = 3; // Direction to follow in
-    float responsiveness = 4 [(mavsdk.options.default_value)="0.5"]; // How responsive the vehicle is to the motion of the target (range 0.0 to 1.0)
+    FollowPerspective follow_perspective = 0; // Perspective to follow
```
JonasVautherin, line 48: It should stay `= 3`. The number here is for protobuf: it is not the default value of the enum :-)

JonasVautherin (changes requested):
Thanks for the contribution!

However, I am not completely sure about renaming everything here. The problem being that it will break the API: whoever updates MAVSDK after that will have to change their code to fix the build. Is that really necessary?

Or maybe could we keep the current naming (even if PX4 renamed "direction" into "perspective"), and just add the new values to the enum (e.g. `FOLLOW_DIRECTION_BEHIND RIGHT` would be a new one consistent with the current naming, without breaking the API?

What do you think?

protos/follow_me/follow_me.proto
```diff
@@ -30,19 +30,23 @@ service FollowMeService {
...
+    enum FollowPerspective {
+        FOLLOW_PERSPECTIVE_BEHIND = 0;      // Follow from behind
+        FOLLOW_PERSPECTIVE_FRONT = 1;       // Follow from front
+        FOLLOW_PERSPECTIVE_FRONTRIGHT = 2;  // Follow from front right
+        FOLLOW_PERSPECTIVE_FRONTLEFT = 3;   // Follow from front left
+        FOLLOW_PERSPECTIVE_RIGHT = 4;       // Follow from right
+        FOLLOW_PERSPECTIVE_LEFT = 5;        // Follow from left
+        FOLLOW_PERSPECTIVE_BEHINDRIGHT = 6; // Follow from behind right
+        FOLLOW_PERSPECTIVE_BEHINDLEFT = 7 ; // Follow from behind left
+        FOLLOW_PERSPECTIVE_INVALID = 8;     // Follow perspective invalid
```
julianoes, line 43: Adding fields is fine, changing them is a bit more tricky, and - as @JonasVautherin said - probably not worth the trouble.

protos/follow_me/follow_me.proto
julianoes: I would leave the variable name but change the comment to make it clearer.

junwoo091400:
Force pushed to apply suggestion comments!

protos/follow_me/follow_me.proto
```diff
@@ -31,18 +31,23 @@ service FollowMeService {
...
-        FOLLOW_DIRECTION_FRONT_LEFT = 4; // Follow from front left
+        FOLLOW_DIRECTION_BEHIND = 0; // Follow from behind
+        FOLLOW_DIRECTION_FRONT = 1; // Follow from front
+        FOLLOW_DIRECTION_FRONT_RIGHT = 2; // Follow from front right
+        FOLLOW_DIRECTION_FRONT_LEFT = 3; // Follow from front left
+        FOLLOW_DIRECTION_RIGHT = 4; // Follow from right
+        FOLLOW_DIRECTION_LEFT = 5; // Follow from left
+        FOLLOW_DIRECTION_BEHIND_RIGHT = 6; // Follow from behind right
+        FOLLOW_DIRECTION_BEHIND_LEFT = 7; // Follow from behind left
+        FOLLOW_DIRECTION_NONE = 8; // Follow direction invalid (defaults to Behind)
```
JonasVautherin, line 44: Changing the indices will have weird protobuf consequences :innocent:. Again those indices are exclusively reserved to the proto format, and don't compare to an enum value in a language like C++.

```suggestion
        FOLLOW_DIRECTION_NONE = 0; // Follow direction invalid (defaults to Behind)
        FOLLOW_DIRECTION_BEHIND = 1; // Follow from behind
        FOLLOW_DIRECTION_FRONT = 2; // Follow from front
        FOLLOW_DIRECTION_FRONT_RIGHT = 3; // Follow from front right
        FOLLOW_DIRECTION_FRONT_LEFT = 4; // Follow from front left
        FOLLOW_DIRECTION_RIGHT = 5; // Follow from right
        FOLLOW_DIRECTION_LEFT = 6; // Follow from left
        FOLLOW_DIRECTION_BEHIND_RIGHT = 7; // Follow from behind right
        FOLLOW_DIRECTION_BEHIND_LEFT = 8; // Follow from behind left
```

JonasVautherin (changes requested):

protos/follow_me/follow_me.proto
junwoo091400: Alright! But how do we deal with the enum values not matching the expected values from the Autopilot then?

https://github.com/mavlink/MAVSDK/blob/main/src/mavsdk/plugins/follow_me/follow_me_impl.cpp#L99

For example here, the enum class gets directly casted into integer & gets sent to the Autopilot. How can I adapt current changes to this code then? Doing Switch case and sending different enum values? (since they will have different values assigned for each follow direction :thinking: )

protos/follow_me/follow_me.proto
```diff
@@ -31,18 +31,25 @@ service FollowMeService {
...
 
-    float min_height_m = 1 [(mavsdk.options.default_value)="8.0"]; // Minimum height for the vehicle in meters (recommended minimum 8 meters)
-    float follow_distance_m = 2 [(mavsdk.options.default_value)="8.0"]; // Distance from target for vehicle to follow in meters (recommended minimum 1 meter)
-    FollowDirection follow_direction = 3; // Direction to follow in
-    float responsiveness = 4 [(mavsdk.options.default_value)="0.5"]; // How responsive the vehicle is to the motion of the target (range 0.0 to 1.0)
+    float min_height_m = 1 [(mavsdk.options.default_value)="8.0"]; // Follow height in meters (recommended minimum 8 meters)
+    float follow_distance_m = 2 [(mavsdk.options.default_value)="8.0"]; // Follow distance to target in meters (recommended minimum 4 meter)
+    FollowDirection follow_direction = 3; // Direction to follow in (FollowPerspective in PX4-Autopilot)
+    float responsiveness = 4 [(mavsdk.options.default_value)="0.1"]; // How responsive the vehicle is to the motion of the target, Lower value = More responsive (range 0.0 to 1.0)
+    float altitude_mode = 5 [(mavsdk.options.default_value)="0"]; // Follow Altitude control mode
```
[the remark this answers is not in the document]
junwoo091400: I have added the enum in the recent commit :+1:  : https://github.com/mavlink/MAVSDK-Proto/pull/287/commits/51bb85756bc130f6d7ae0be8f93fd0545d2df039

protos/follow_me/follow_me.proto
```diff
@@ -30,19 +30,32 @@ service FollowMeService {
...
 
-    float min_height_m = 1 [(mavsdk.options.default_value)="8.0"]; // Minimum height for the vehicle in meters (recommended minimum 8 meters)
-    float follow_distance_m = 2 [(mavsdk.options.default_value)="8.0"]; // Distance from target for vehicle to follow in meters (recommended minimum 1 meter)
-    FollowDirection follow_direction = 3; // Direction to follow in
-    float responsiveness = 4 [(mavsdk.options.default_value)="0.5"]; // How responsive the vehicle is to the motion of the target (range 0.0 to 1.0)
+    // Altitude mode to configure which altitude the follow me will assume the target to be at.
+    enum FollowAltitudeMode {
+        FOLLOW_ALTITUDE_MODE_CONSTANT = 0; // Target assumed to be mobing at a constant altitude of home position (where the vehicle armed)
+        FOLLOW_ALTITUDE_MODE_TERRAIN = 1; // Target assumed to be at the terrain level sensed by the distance sensor
+        FOLLOW_ALTITUDE_MODE_TRACKTARGET = 2; // Target GPS altitude taken into account to do 3D tracking
```
JonasVautherin, line 50: I would write it like this:

```suggestion
        FOLLOW_ALTITUDE_MODE_TRACK_TARGET = 2; // Target GPS altitude taken into account to do 3D tracking
```

So that it becomes `trackTarget` or `track_target` or `TRACK_TARGET` in the generated codes for the language bindings (mavsdk-python, mavsdk-java, etc).

protos/follow_me/follow_me.proto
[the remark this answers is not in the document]
JonasVautherin: Nice! :blush:

protos/follow_me/follow_me.proto
junwoo091400: Gotcha, wasn't aware of that! Protobuf stuff :mask:

protos/follow_me/follow_me.proto
junwoo091400: Ohhhh so technically all the code in MAVSDK that has those enums must do a switch case to manually map out the actual values! Then this follow me code was initially dangerous :smirk:

junwoo091400:
Rebased and added the latest commit on depreciating follow_direction!

protos/follow_me/follow_me.proto
```diff
@@ -30,19 +30,42 @@ service FollowMeService {
...
+        FOLLOW_ALTITUDE_MODE_TERRAIN = 1; // Target assumed to be at the terrain level sensed by the distance sensor
+        FOLLOW_ALTITUDE_MODE_TARGET_GPS = 2; // Target GPS altitude taken into account to do 3D tracking
+    }
+
+    float min_height_m = 1 [(mavsdk.options.default_value)="8.0"]; // Follow height in meters (recommended minimum 8 meters)
+    float follow_distance_m = 2 [(mavsdk.options.default_value)="8.0"]; // Follow distance to target in meters (recommended minimum 4 meter)
+    FollowDirection follow_direction = 3; // [!DEPRECIATED!] Direction to follow target at. Exists as a legacy to old PX4 Follow-Me, but now only used for setting the follow_angle_deg. This value never gets sent to the PX4 side!
+    float responsiveness = 4 [(mavsdk.options.default_value)="0.1"]; // How responsive the vehicle is to the motion of the target, Lower value = More responsive (range 0.0 to 1.0)
+    FollowAltitudeMode altitude_mode = 5; // Follow Altitude control mode
+    float max_tangential_vel = 6 [(mavsdk.options.default_value)="8.0"]; // Maximum orbit tangential velocity relative to the target. Higher value = More aggressive follow angle tracking.
```
JonasVautherin, line 67: Just realized: we add the units to the variable name (e.g. [see here](https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/telemetry/telemetry.proto#L546)).

```suggestion
    float max_tangential_vel_m_s = 6 [(mavsdk.options.default_value)="8.0"]; // Maximum orbit tangential velocity relative to the target, in meters per second. Higher value = More aggressive follow angle tracking.
```

protos/follow_me/follow_me.proto
```diff
@@ -30,19 +30,42 @@ service FollowMeService {
...
-    // Direction relative to the target that the vehicle should follow
+    /** [!DEPRECIATED!]
+      * Direction relative to the target that the vehicle should follow.
+      * Measured in reference to target's course (where it's moving towards)
+      * 
+      * It is only used for setting the follow angle indirectly, for legacy reasons
+      * (since PX4 Follow Me used to use FollowDirection enums for setting, but now
+      * uses the follow_angle_deg float value).
+      * 
+      * DO NOT USE if possible!
```
julianoes, line 41: @junwoo091400 so what's the new API? Just an angle to define this? And if so, would it make sense to through this API out and have the new one with MAVSDK v2 which is probably going to be released in the next weeks anyway?

junwoo091400: The new API is to just setting the angle in degrees via the `FLW_TGT_FA` parameter. If removing this `follow_direction` is ok, I would love to! (Since it requires having switch cases, etc (Having enums is painful :rofl:)

Do I just drop it in this PR then?

JonasVautherin: Yeah, just drop it :+1:.

junwoo091400:
Just to be clear, can I also drop the `follow_direction` in the .proto file? I remember that @JonasVautherin you mentioned that:
1. Renaming
2. Changing proto value

is not good (in Protobuf perspective). Does that mean I can't even remove the variable? Or, can I remove it but I just need to keep the values for the other variables constant? (that is, just jump the case  `FollowDirection follow_direction = 3` and omit the number 3)

JonasVautherin:
> I remember that @JonasVautherin you mentioned that

Yeah that's right, but I had not realized that we were gonna go to MAVSDK v2.x.x soon. Because it's a major update, we can break the API (that's what the major version number means :innocent:).

So exceptionally (because we had to break the API and therefore move to v2 soon), I guess feel free to rename everything you want here :rofl:.

protos/follow_me/follow_me.proto
```diff
@@ -59,13 +38,13 @@ message Config {
         FOLLOW_ALTITUDE_MODE_TARGET_GPS = 2; // Target GPS altitude taken into account to do 3D tracking
     }
 
-    float min_height_m = 1 [(mavsdk.options.default_value)="8.0"]; // Follow height in meters (recommended minimum 8 meters)
-    float follow_distance_m = 2 [(mavsdk.options.default_value)="8.0"]; // Follow distance to target in meters (recommended minimum 4 meter)
-    FollowDirection follow_direction = 3; // [!DEPRECIATED!] Direction to follow target at. Exists as a legacy to old PX4 Follow-Me, but now only used for setting the follow_angle_deg. This value never gets sent to the PX4 side!
+    float min_height_m = 1 [(mavsdk.options.default_value)="8.0"]; // [m] Follow height in meters (recommended minimum 8 meters)
+    float follow_distance_m = 2 [(mavsdk.options.default_value)="8.0"]; // [m] Follow distance to target in meters (recommended minimum 4 meter)
+    // FollowDirection follow_direction = 3; // [!DEPRECIATED!] Direction to follow target at. Exists as a legacy to old PX4 Follow-Me, but now only used for setting the follow_angle_deg. This value never gets sent to the PX4 side!
```
JonasVautherin, line 43: Yeah makes sense to keep it commented out, so that we know that we should not reuse id 3!

JonasVautherin (approved):
Fine for me, with the breaking API change going into MAVSDK v2 :+1:

julianoes (dismissed):

junwoo091400:
Thanks for the approval! can this get merged? :pray:

JonasVautherin:
> Thanks for the approval! can this get merged?

I think we first need the corresponding PX4 PR to get merged. Right now this doesn't work against PX4 upstream...

julianoes:
> The PR is currently having some build target flash overflow issues, which is why the merge is being delayed

Of course :(.

junwoo091400:
Rebased on `main` branch

junwoo091400:
Squashed to make into a single commit

junwoo091400:
Good for merge as FollowMe was merged into master as well 👍

PR #290, opened 2022-06-02
Battery message - extend with status, current, and fix percentage

hamishwillee:
This is proposed proto to match the updated battery status updates in https://github.com/mavlink/MAVSDK/pull/1792

It fixes the percentage remaining to be 0-100 (docs only) and adds current and status flags as suggested by @julianoes 

However, in draft because the status flags are still in flux. This shouldn't be merged!

But would like review to sanity check it isn't wrong/broken (then I can try next update step)

### DIFF
--- a/protos/telemetry/telemetry.proto
+++ b/protos/telemetry/telemetry.proto
@@ -564,11 +564,37 @@ enum FixType {
     FIX_TYPE_RTK_FIXED = 6; // RTK Fixed, 3D position
 }
 
+
+enum BatteryStatusFlags {
+    BATTERY_STATUS_FLAGS_READY_TO_USE = 1; // The battery is ready to use (fly).
+    BATTERY_STATUS_FLAGS_CHARGING = 2; // Battery is charging.
+    BATTERY_STATUS_FLAGS_CELL_BALANCING = 4; // Battery is cell balancing (during charging).
+    BATTERY_STATUS_FLAGS_FAULT_CELL_IMBALANCE = 8; // Battery cells are not balanced.
+    BATTERY_STATUS_FLAGS_AUTO_DISCHARGING = 16; // Battery is auto discharging (towards storage level).
+    BATTERY_STATUS_FLAGS_REQUIRES_SERVICE = 32; // Battery requires service (not safe to fly). 
+    BATTERY_STATUS_FLAGS_BAD_BATTERY = 64; // Battery is faulty and cannot be repaired (not safe to fly).
+    BATTERY_STATUS_FLAGS_PROTECTIONS_ENABLED = 128; // Automatic battery protection monitoring is enabled. 
+    BATTERY_STATUS_FLAGS_FAULT_PROTECTION_SYSTEM = 256; // The battery fault protection system had detected a fault and cut all power from the battery.
+    BATTERY_STATUS_FLAGS_FAULT_OVER_VOLT = 512; // One or more cells are above their maximum voltage rating.
+    BATTERY_STATUS_FLAGS_FAULT_UNDER_VOLT = 1024; // One or more cells are below their minimum voltage rating.
+    BATTERY_STATUS_FLAGS_FAULT_OVER_TEMPERATURE = 2048; // Over-temperature fault. 
+    BATTERY_STATUS_FLAGS_FAULT_UNDER_TEMPERATURE = 4096; // Under-temperature fault.
+    BATTERY_STATUS_FLAGS_FAULT_OVER_CURRENT = 8192; // Over-current fault.
+    BATTERY_STATUS_FLAGS_FAULT_SHORT_CIRCUIT = 16384; // Short circuit event detected.
+    BATTERY_STATUS_FLAGS_FAULT_INCOMPATIBLE_VOLTAGE = 32768; // Voltage not compatible with power rail voltage (batteries on same power rail should have similar voltage).
+    BATTERY_STATUS_FLAGS_FAULT_INCOMPATIBLE_FIRMWARE = 65536; // Battery firmware is not compatible with current autopilot firmware.
+    BATTERY_STATUS_FLAGS_FAULT_INCOMPATIBLE_CELLS_CONFIGURATION = 131072; // Battery is not compatible due to cell configuration.
+    BATTERY_STATUS_FLAGS_CAPACITY_RELATIVE_TO_FULL = 262144; // Battery capacity values are relative to a known-full battery (smart battery) and not estimated.
+    BATTERY_STATUS_FLAGS_EXTENDED = 4294967295; // Reserved for future use.
+}
+
 // Battery type.
 message Battery {
     uint32 id = 3 [(mavsdk.options.default_value)="0"]; // Battery ID, for systems with multiple batteries
     float voltage_v = 1 [(mavsdk.options.default_value)="NaN"]; // Voltage in volts
-    float remaining_percent = 2 [(mavsdk.options.default_value)="NaN"]; // Estimated battery remaining (range: 0.0 to 1.0)
+    float remaining_percent = 2 [(mavsdk.options.default_value)="NaN"]; // Estimated battery remaining (range: 0 to 100)
+    float current_ma = 4 [(mavsdk.options.default_value)="NaN"]; // Current in mA
+    BatteryStatusFlags status = 5; // Battery status flags
 }
 
 /*


protos/telemetry/telemetry.proto
```diff
@@ -564,11 +564,37 @@ enum FixType {
...
+    BATTERY_STATUS_FLAGS_FAULT_UNDER_VOLT = 1024; // One or more cells are below their minimum voltage rating.
+    BATTERY_STATUS_FLAGS_FAULT_OVER_TEMPERATURE = 2048; // Over-temperature fault. 
+    BATTERY_STATUS_FLAGS_FAULT_UNDER_TEMPERATURE = 4096; // Under-temperature fault.
+    BATTERY_STATUS_FLAGS_FAULT_OVER_CURRENT = 8192; // Over-current fault.
+    BATTERY_STATUS_FLAGS_FAULT_SHORT_CIRCUIT = 16384; // Short circuit event detected.
+    BATTERY_STATUS_FLAGS_FAULT_INCOMPATIBLE_VOLTAGE = 32768; // Voltage not compatible with power rail voltage (batteries on same power rail should have similar voltage).
+    BATTERY_STATUS_FLAGS_FAULT_INCOMPATIBLE_FIRMWARE = 65536; // Battery firmware is not compatible with current autopilot firmware.
+    BATTERY_STATUS_FLAGS_FAULT_INCOMPATIBLE_CELLS_CONFIGURATION = 131072; // Battery is not compatible due to cell configuration.
+    BATTERY_STATUS_FLAGS_CAPACITY_RELATIVE_TO_FULL = 262144; // Battery capacity values are relative to a known-full battery (smart battery) and not estimated.
+    BATTERY_STATUS_FLAGS_EXTENDED = 4294967295; // Reserved for future use.
```
JonasVautherin, line 588: This ID number is too high for protobuf (hence the prototool failure in the CI) :grin:

JonasVautherin (changes requested):

protos/telemetry/telemetry.proto
julianoes: We generally don't support enums used as bitflags in MAVSDK, right @JonasVautherin?

So what we can do is add boolean values for it and then convert between the flags and bools.
I think this makes using the API easier for other languages, where it's not as common to do bitwise checks.

protos/telemetry/telemetry.proto
```diff
@@ -564,11 +564,37 @@ enum FixType {
...
+}
+
 // Battery type.
 message Battery {
     uint32 id = 3 [(mavsdk.options.default_value)="0"]; // Battery ID, for systems with multiple batteries
     float voltage_v = 1 [(mavsdk.options.default_value)="NaN"]; // Voltage in volts
-    float remaining_percent = 2 [(mavsdk.options.default_value)="NaN"]; // Estimated battery remaining (range: 0.0 to 1.0)
+    float remaining_percent = 2 [(mavsdk.options.default_value)="NaN"]; // Estimated battery remaining (range: 0 to 100)
+    float current_ma = 4 [(mavsdk.options.default_value)="NaN"]; // Current in mA
```
julianoes, line 596: Could also be `A` instead of `mA` but then we are quite different from the MAVLink message.

protos/telemetry/telemetry.proto
JonasVautherin: Well as long as the number is in bounds, it's fine. Usually we just increment from 0, but well... I don't really mind if we want to use weird ids :sweat_smile:.

However the protobuf field number is only a protobuf thing, it's completely independent from MAVLink. So using powers of two has no meaning here :+1:

julianoes: So that's the thing, we don't support it because the numbers are removed in the C++ API.

JonasVautherin: Sorry I did not get that. In the C++ implementation, the code will have to translate this enum into the mavlink message. But using powers of 2 here doesn't do anything, so it would probably be cleaner to just increment from 0.

protos/telemetry/telemetry.proto
hamishwillee: I can change if you want. Just let me know.

protos/telemetry/telemetry.proto
hamishwillee: I thought I was defining an enum with preset values, but it sounds like you're saying that the value I am assigning in (say) ` BATTERY_STATUS_FLAGS_FAULT_OVER_CURRENT = 8192;` is a protobuf field number.

So what the hell is the resulting enum value, and how would I use it in a bitmask? My best-guess would be that (if the number is just an internal protobuf thing) the enums would copy across to whatever language in whatever form it understands as an auto-numbered enum from zero. 
The implication being that you CANT use these as flags right?

So what I "think" you (@julianoes) are saying is that I can't expose enums as bitmaps through the API. I need to create bools for each flag. Right?
So in the proto I'd get rid of these enums and it might look like:

```
bool status_not_ready_to_use= 5  [(mavsdk.options.default_value)="0"]; // Battery status: not ready to use
bool status_charging= 6  [(mavsdk.options.default_value)="0"]; // Battery status: charging
...
````

I'm right when I assume that these definitions aren't described in the docs right? (checked via search, perhaps I just missed it)?

Comment (JonasVautherin) [reply protos/telemetry/telemetry.proto:588]: > So what the hell is the resulting enum value, and how would I use it in a bitmask?

That's an implementation detail, that's not the responsibility of the API. Using bitmasks is an optimization at the protocol level, it has nothing to do here.

The part of the code that will create and send the actual mavlink message will understand the enum, and convert it to whatever mavlink wants :slightly_smiling_face:.

> I need to create bools for each flag. Right?

No, the enum is fine. As a MAVSDK user, I'd rather deal with `BATTERY_STATUS_FLAGS_EXTENDED` than `4294967295`. I don't care how the enum is encoded on the wire, be it with a bitmask, with booleans, little-endian or not, UTF8 or not, etc.

> I'm right when I assume that these definitions aren't described in the docs right?

The proto definition is documented e.g. [here](https://developers.google.com/protocol-buffers/docs/proto3) if you are interested. But let's be honest, nobody ever reads that. Instead they look at the other MAVSDK protos, write their version, and let us maintainers tell them if that's okay or not.

Also I believe that if you tried to generate the code and implement the C++ code, you would see that this API gives you an enum and that you need to translate it where you create the mavlink message :innocent:.

Comment (julianoes) [reply protos/telemetry/telemetry.proto:588]: > No, the enum is fine. As a MAVSDK user, I'd rather deal with BATTERY_STATUS_FLAGS_EXTENDED than 4294967295. I don't care how the enum is encoded on the wire, be it with a bitmask, with booleans, little-endian or not, UTF8 or not, etc.

@JonasVautherin I don't think this would work. The problem is that the MAVLink spec is not really an enum but it's bitflags. 

So in our conversion the enum becomes a normal enum incremented with 1 (instead of doubled for bits), so you can only ever set one flag at a time. In the MAVLink world you can set multiple bits.

Therefore, I'm suggesting to use boolean flags for each flag, and then encode it to bits internally.

Comment (JonasVautherin) [reply protos/telemetry/telemetry.proto:588]: Ooooooh, I get it now :+1:. Yeah then a boolean makes total sense :100:

Comment (hamishwillee) [reply protos/telemetry/telemetry.proto:588]: So is the format I suggested right?

```
bool status_not_ready_to_use= 5  [(mavsdk.options.default_value)="0"]; // Battery status: not ready to use
bool status_charging= 6  [(mavsdk.options.default_value)="0"]; // Battery status: charging
...
```

File: protos/telemetry/telemetry.proto [hunk 3]
```diff
@@ -564,13 +564,38 @@ enum FixType {
...
+    bool status_not_ready_to_use = 5  [(mavsdk.options.default_value)="0"]; // Battery not ready to use
+    bool status_charging = 6  [(mavsdk.options.default_value)="0"]; // Battery charging
+    
+    bool status_cell_balancing = 7  [(mavsdk.options.default_value)="0"]; // Battery is cell balancing (during charging)
```
hamishwillee, line 579: actually, here they all are, so you can tell me if I did it right :-)
These are still all in discussion

protos/telemetry/telemetry.proto
JonasVautherin: I would try with `[(mavsdk.options.default_value)="false"]`, I think that should work :innocent:

julianoes: Yes, that looks right.

protos/telemetry/telemetry.proto
```diff
@@ -564,13 +564,38 @@ enum FixType {
...
+    bool battery_fault_over_temperature = 16  [(mavsdk.options.default_value)="0"]; // Over-temperature fault
+    bool battery_fault_under_temperature = 17  [(mavsdk.options.default_value)="0"]; // Under-temperature fault
+    bool battery_fault_over_current = 18 [(mavsdk.options.default_value)="0"]; // Over-current fault
+    bool battery_fault_short_curcuit = 19  [(mavsdk.options.default_value)="0"]; // Short circuit event detected
+    bool battery_fault_incompatible_voltage = 20  [(mavsdk.options.default_value)="0"]; // Voltage not compatible with power rail voltage (batteries on same power rail should have similar voltage)
+    bool battery_fault_incompatible_firmware = 21  [(mavsdk.options.default_value)="0"]; // Battery firmware is not compatible with current autopilot firmware
+    bool battery_fault_incompatible_cells_configuration = 22  [(mavsdk.options.default_value)="0"]; // Battery is not compatible due to cell configuration
+    bool status_capacity_relative_to_full = 23  [(mavsdk.options.default_value)="0"]; // Battery capacity values are relative to a known-full battery (smart battery) and not estimated
```
hamishwillee, line 595:
```suggestion
    bool status_not_ready_to_use = 5  [(mavsdk.options.default_value)="false"]; // Battery not ready to use
    bool status_charging = 6  [(mavsdk.options.default_value)="false"]; // Battery charging
    bool status_cell_balancing = 7  [(mavsdk.options.default_value)="false"]; // Battery is cell balancing (during charging)
    bool battery_fault_cells_not_balanced = 8  [(mavsdk.options.default_value)="false"]; // Cells are not balanced
    bool status_auto_discharging = 9  [(mavsdk.options.default_value)="false"]; // Battery is auto discharging (towards storage level)
    bool status_requires_service = 10  [(mavsdk.options.default_value)="false"]; // Battery requires service (not safe to fly)
    bool status_bad_battery = 11 [(mavsdk.options.default_value)="false"]; // Battery is faulty and cannot be repaired (not safe to fly)
    bool status_protections_enabled = 12  [(mavsdk.options.default_value)="false"]; // Automatic battery protection monitoring is enabled. 
    bool battery_fault_protection_system = 13  [(mavsdk.options.default_value)="false"]; // The battery fault protection system had detected a fault and cut all power from the battery
    bool battery_fault_over_volt = 14  [(mavsdk.options.default_value)="false"]; // One or more cells are above their maximum voltage rating
    bool battery_fault_under_volt = 15  [(mavsdk.options.default_value)="false"]; // One or more cells are below their minimum voltage rating
    bool battery_fault_over_temperature = 16  [(mavsdk.options.default_value)="false"]; // Over-temperature fault
    bool battery_fault_under_temperature = 17  [(mavsdk.options.default_value)="false"]; // Under-temperature fault
    bool battery_fault_over_current = 18 [(mavsdk.options.default_value)="false"]; // Over-current fault
    bool battery_fault_short_curcuit = 19  [(mavsdk.options.default_value)="false"]; // Short circuit event detected
    bool battery_fault_incompatible_voltage = 20  [(mavsdk.options.default_value)="false"]; // Voltage not compatible with power rail voltage (batteries on same power rail should have similar voltage)
    bool battery_fault_incompatible_firmware = 21  [(mavsdk.options.default_value)="false"]; // Battery firmware is not compatible with current autopilot firmware
    bool battery_fault_incompatible_cells_configuration = 22  [(mavsdk.options.default_value)="false"]; // Battery is not compatible due to cell configuration
    bool status_capacity_relative_to_full = 23  [(mavsdk.options.default_value)="false"]; // Battery capacity values are relative to a known-full battery (smart battery) and not estimated
```

julianoes:
@hamishwillee what's the status here? Is this good to go in?

hamishwillee:
@julianoes Ha ha ha ha :-)

No. The battery messages are still a little in flux. Since MAVSDK takes common rather than development, I don't want this in yet,
It would be good if, independent of this, the remaining percentage changing to a percent value could go in.

julianoes:
@hamishwillee FYI the v1 PR is in.

PR #291, opened 2022-06-21
Transponder: use altitude type to be more accurate

chep:
There are 2 types of altitude: from baro source (AMSL) and from gnss source (using geoid)

We need to inform user of what type is used because sometimes there is a huge difference.


protos/transponder/transponder.proto
```diff
@@ -53,19 +53,26 @@ enum AdsbEmitterType {
...
+}
+
+
 // ADSB Vehicle type.
 message AdsbVehicle {
     uint32 icao_address = 1; // ICAO (International Civil Aviation Organization) unique worldwide identifier
     double latitude_deg = 2; // Latitude in degrees (range: -90 to +90)
     double longitude_deg = 3; // Longitude in degrees (range: -180 to +180).
-    float absolute_altitude_m = 5; // Altitude AMSL (above mean sea level) in metres
+    float altitude_m = 5; // Altitude ASL (above sea level) in metres
```
JonasVautherin, line 67: Do we really need to rename that field? That will break backward compatibility.

Keeping the old name and using a default value for `altitude_type` in the C++ implementation should stay backward compatible, which would be nicer, I think.

JonasVautherin (approved):

julianoes (commented):
I think this is fine. Do we generally have only one of the two available, or also both?

chep:
> I think this is fine. Do we generally have only one of the two available, or also both?

We have only one altitude type in the mavlink message. I think if a device manages two types it needs to send 2 messages. But I don't think this kind of device exists.

julianoes:
Ok, in that case I'd say this is fine.

JonasVautherin (changes requested):

JonasVautherin:
FYI @chep: the MAVSDK-C++ PR seems fine, I'll merge it as soon as this one is finished and merged.

chep:
I just added comments and changed `AdsbAltitudeType altitude_type = 15` by `AdsbAltitudeType altitude_type = 4` because `4` was not used and it is the same order than mavlink message: https://mavlink.io/en/messages/common.html#ADSB_VEHICLE

JonasVautherin (approved):

PR #296, opened 2022-09-10
param: allow to set component ID and protocol

julianoes:
This should allow to create param plugins to talk to specific components.

This replaces the `param->late_init()` suggested in https://github.com/mavlink/MAVSDK/pull/1809 by @Consti10.


JonasVautherin (approved):

JonasVautherin:
Now I'm wondering: what happens if I create a `param` plugin, use it with the autopilot for a while, and then select another component? Will I still receive update for autopilot updates?

I suppose it would be better to be able to specify the component in the constructor (forcing one param instance per remote component), but we don't have that (yet) in the language wrappers :thinking:.

julianoes:
Good question. I don't know. Adding args to the ctor would be nice indeed.

JonasVautherin (approved):

PR #297, opened 2022-09-13
PR Add battery proto status message for v1

Katawann:
I am interested in receiving more information about the battery status through MAVSDK. Mainly I need the current consumption, but I took the opportunity to add the other information.
   - temperature
   - current_battery
   - current_consumed
   - energy_consumed

I don't see the _temperature_ and _energy_consumed_ in my flight logs so not sure if it is relevant to add them. I also saw that there was already a PR in progress for adding messages for the v2 battery but nothing done for v1 from what I found.

### DIFF
--- a/protos/telemetry/telemetry.proto
+++ b/protos/telemetry/telemetry.proto
@@ -593,9 +593,12 @@ enum FixType {
 
 // Battery type.
 message Battery {
-    uint32 id = 3 [(mavsdk.options.default_value)="0"]; // Battery ID, for systems with multiple batteries
-    float voltage_v = 1 [(mavsdk.options.default_value)="NaN"]; // Voltage in volts
-    float remaining_percent = 2 [(mavsdk.options.default_value)="NaN"]; // Estimated battery remaining (range: 0.0 to 1.0)
+    uint32 id = 1 [(mavsdk.options.default_value)="0"]; // Battery ID, for systems with multiple batteries
+    float temperature_degc = 2 [(mavsdk.options.default_value)="NaN"]; // Temperature of the battery in degrees Celsius. NAN for unknown temperature
+    float voltage_v = 3 [(mavsdk.options.default_value)="NaN"]; // Voltage in volts
+    float current_battery_a = 4 [(mavsdk.options.default_value)="NaN"]; // Battery current in Amps, NAN if autopilot does not measure the current
+    float capacity_consumed_ah = 5 [(mavsdk.options.default_value)="NaN"]; // Consumed charge in Amp hours, NAN if autopilot does not provide consumption estimate
+    float remaining_percent = 6 [(mavsdk.options.default_value)="NaN"]; // Estimated battery remaining (range: 0 to 100)
 }
 
 /*


protos/telemetry/telemetry.proto
```diff
@@ -576,7 +576,11 @@ enum FixType {
 // Battery type.
 message Battery {
     uint32 id = 3 [(mavsdk.options.default_value)="0"]; // Battery ID, for systems with multiple batteries
+    float temperature_cdegC = 7 [(mavsdk.options.default_value)="NaN"]; // Temperature of the battery. INT16_MAX for unknown temperature
```
JonasVautherin, line 579: Why not `degC` instead of `cdegC`?

protos/telemetry/telemetry.proto
```diff
@@ -576,7 +576,11 @@ enum FixType {
...
     float voltage_v = 1 [(mavsdk.options.default_value)="NaN"]; // Voltage in volts
+    float current_battery_cA = 4 [(mavsdk.options.default_value)="NaN"]; // Battery current, -1: autopilot does not measure the current
```
JonasVautherin, line 581: Wouldn't `A` or `mA` be a bit more natural? :sweat_smile:

protos/telemetry/telemetry.proto
Katawann: agree

protos/telemetry/telemetry.proto
Katawann: agree as well. I took the format from the BATTERY_STATUS message but probably not the most meaningful units

protos/telemetry/telemetry.proto
```diff
@@ -576,7 +576,11 @@ enum FixType {
 // Battery type.
 message Battery {
     uint32 id = 3 [(mavsdk.options.default_value)="0"]; // Battery ID, for systems with multiple batteries
+    float temperature_degc = 7 [(mavsdk.options.default_value)="NaN"]; // Temperature of the battery. INT16_MAX for unknown temperature
```
julianoes, line 579: Let's use NAN for unknown.

julianoes: And why is the ordering so odd? I'd probably move the new ones down.

protos/telemetry/telemetry.proto
```diff
@@ -576,7 +576,11 @@ enum FixType {
...
     float voltage_v = 1 [(mavsdk.options.default_value)="NaN"]; // Voltage in volts
+    float current_battery_ma = 4 [(mavsdk.options.default_value)="NaN"]; // Battery current, -1: autopilot does not measure the current
```
julianoes, line 581: NAN for unknown, I'd say. And what about using A?

protos/telemetry/telemetry.proto
```diff
@@ -576,7 +576,11 @@ enum FixType {
...
+    float current_consumed_mah = 5 [(mavsdk.options.default_value)="NaN"]; // Consumed charge, -1: autopilot does not provide consumption estimate
+    float energy_consumed_hj = 6 [(mavsdk.options.default_value)="NaN"]; // Consumed energy, -1: autopilot does not provide energy consumption estimate
```
julianoes, line 583: What's hecto Joule? Can we just make it Joule?

protos/telemetry/telemetry.proto
```diff
@@ -576,7 +576,11 @@ enum FixType {
 // Battery type.
 message Battery {
     uint32 id = 3 [(mavsdk.options.default_value)="0"]; // Battery ID, for systems with multiple batteries
+    float temperature_degc = 7 [(mavsdk.options.default_value)="NaN"]; // Temperature of the battery. INT16_MAX for unknown temperature
     float voltage_v = 1 [(mavsdk.options.default_value)="NaN"]; // Voltage in volts
+    float current_battery_ma = 4 [(mavsdk.options.default_value)="NaN"]; // Battery current, -1: autopilot does not measure the current
+    float current_consumed_mah = 5 [(mavsdk.options.default_value)="NaN"]; // Consumed charge, -1: autopilot does not provide consumption estimate
```
julianoes, line 582: NAN for unknown. And what about using Ah

hamishwillee:
There's only one battery API from a MAVSDK perspective right? If it were me I would make that match the MAVLink v2  in terms of units and naming, where possible. Most of it maps pretty cleanly - its really only the faults and status that are in discussion.

Katawann:
@hamishwillee I saw your [proposition ](https://github.com/mavlink/MAVSDK/pull/1792) for new battery status v2 but where you created a new function so I thought it was you using another mavlink message. So I can bring the modifications you did in #290 without the status information so that I can make it works with v1. What do you think ?

protos/telemetry/telemetry.proto
Katawann: I bring the units from the mavlink message, but I agree that if the end user needs to change the units every time it makes more sense that we do it right away

Katawann:
@julianoes would you have time to review it so that I can integrate the functionality further away ?

protos/telemetry/telemetry.proto
```diff
@@ -577,7 +577,11 @@ enum FixType {
 message Battery {
     uint32 id = 3 [(mavsdk.options.default_value)="0"]; // Battery ID, for systems with multiple batteries
     float voltage_v = 1 [(mavsdk.options.default_value)="NaN"]; // Voltage in volts
-    float remaining_percent = 2 [(mavsdk.options.default_value)="NaN"]; // Estimated battery remaining (range: 0.0 to 1.0)
+    float remaining_percent = 2 [(mavsdk.options.default_value)="NaN"]; // Estimated battery remaining (range: 0 to 100)
+    float current_battery_a = 4 [(mavsdk.options.default_value)="NaN"]; // Battery current, NAN if autopilot does not measure the current
+    float current_consumed_ah = 5 [(mavsdk.options.default_value)="NaN"]; // Consumed charge, NAN if autopilot does not provide consumption estimate
+    float energy_consumed_j = 6 [(mavsdk.options.default_value)="NaN"]; // Consumed energy, NAN if autopilot does not provide energy consumption estimate
+    float temperature_degc = 7 [(mavsdk.options.default_value)="NaN"]; // Temperature of the battery. NAN for unknown temperature
```
julianoes, line 584:
```suggestion
    float current_battery_a = 4 [(mavsdk.options.default_value)="NaN"]; // Battery current in Amps, NAN if autopilot does not measure the current
    float current_consumed_ah = 5 [(mavsdk.options.default_value)="NaN"]; // Consumed charge in Amp hours, NAN if autopilot does not provide consumption estimate
    float energy_consumed_wh = 6 [(mavsdk.options.default_value)="NaN"]; // Consumed energy in Watt hours, NAN if autopilot does not provide energy consumption estimate
    float temperature_degc = 7 [(mavsdk.options.default_value)="NaN"]; // Temperature of the battery in degrees Celsius. NAN for unknown temperature
```

What about that?

Katawann: Thank you for your suggestion, I did the changes

julianoes (dismissed):
This is good for me. @hamishwillee do you want to veto this or are you ok with it? I don't understand all the things going on with battery, so I'm asking you.

protos/telemetry/telemetry.proto
```diff
@@ -577,7 +577,11 @@ enum FixType {
 message Battery {
     uint32 id = 3 [(mavsdk.options.default_value)="0"]; // Battery ID, for systems with multiple batteries
     float voltage_v = 1 [(mavsdk.options.default_value)="NaN"]; // Voltage in volts
-    float remaining_percent = 2 [(mavsdk.options.default_value)="NaN"]; // Estimated battery remaining (range: 0.0 to 1.0)
+    float remaining_percent = 2 [(mavsdk.options.default_value)="NaN"]; // Estimated battery remaining (range: 0 to 100)
+    float current_battery_a = 4 [(mavsdk.options.default_value)="NaN"]; // Battery current in Amps, NAN if autopilot does not measure the current
```
hamishwillee, line 581: @julianoes The battery message itself will be a value in mA. If the conversion doesn't matter I'm fine with this being in Amps.

protos/telemetry/telemetry.proto
```diff
@@ -577,7 +577,11 @@ enum FixType {
...
+    float current_consumed_ah = 5 [(mavsdk.options.default_value)="NaN"]; // Consumed charge in Amp hours, NAN if autopilot does not provide consumption estimate
```
hamishwillee, line 582: The battery message will be in mA hours. Same comment as for the current.

protos/telemetry/telemetry.proto
```diff
@@ -577,7 +577,11 @@ enum FixType {
...
+    float energy_consumed_wh = 6 [(mavsdk.options.default_value)="NaN"]; // Consumed energy in Watt hours, NAN if autopilot does not provide energy consumption estimate
```
hamishwillee, line 583: From a MAVLink point of view this duplicates the field above (we don't send both, and mAh is a much more common representation in drones). If MAVSDK chooses to provide the conversion to make it easier for consumers I have no problem with it.

protos/telemetry/telemetry.proto
hamishwillee: Note also that in the new battery message there will be a status flag MAV_BATTERY_STATUS_FLAGS_CAPACITY_RELATIVE_TO_FULL that will _change/clarify the meaning of this field_. Currently the assumption is that this is the consumed current measured by a power module; you start with a full battery on power on, and this goes up, the remaining capacity goes down. In the new version, the flag will tell you that the capacity consumed (note, not "current consumed") that this value is from a smart battery and is expected to be "amount to full battery" not "based on current since vehicle powered".

It might be that you want to think about exposing an API that separates these - i.e. you might have current consumed only populated if MAV_BATTERY_STATUS_FLAGS_CAPACITY_RELATIVE_TO_FULL is not set, and a separate one for capacity in the next version.

hamishwillee (dismissed):
@julianoes I'm OK with this. I've added some comments in case you would prefer to match MAVLink - i.e. MAVLink sends mA rather than A for efficiency reasons, but I am not sure whether consumers care.

I think the new battery message will be coming out in the next couple of days. We're in a final broader-round review stage. The main additions that will come from that will be the status fields.
The most interesting change is covered in https://github.com/mavlink/MAVSDK-Proto/pull/297/files#r987400519 - but I think that if you leave things exactly as they are you can then simply extend the MAVSDK with capacity consumed and capacity remaining in mAh that are populated only when the relative to full flag is set.

protos/telemetry/telemetry.proto
julianoes: You're referring to the mAh capacity of a battery. But here we are talking about Amps drawn from it, right?

protos/telemetry/telemetry.proto
julianoes: Consumed is generally in mAh, I can see that point. So now the question is whether we go SI as much as possible, or units that people are familiar with.

protos/telemetry/telemetry.proto
julianoes: It's not the same though. Amp hours is the Amps drawn over time. If you also have the voltage, then you can get to energy consumed.

The MAVLink message seems to have both actually:
https://mavlink.io/en/messages/common.html#BATTERY_STATUS

julianoes:
So what do we do to this MAVSDK API once the new message is out? I would like not to carry two different messages. The point of MAVSDK is to avoid that sort of mess.

Katawann:
> I think the new battery message will be coming out in the next couple of days. We're in a final broader-round review stage. The main additions that will come from that will be the status fields.

If @hamishwillee is close to get something with v2 I guess it's worth waiting the end of the review stage to ensure we have similar messages for non smart battery communication

protos/telemetry/telemetry.proto
hamishwillee: @julianoes No one uses that in drones, and it is not in the new MAVLink message. But if you can derive it from the data that is provided, then sure.

Docs not live yet, but it is in development.xml here: https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/development.xml#L410-L425

hamishwillee:
The latest version of the battery_status_v2 just went into development.xml. This is compatible/good. If it were me I'd remove energy consumed because anecdotally it is rarely reported by a GCS - and it definitely won't be sent in the new battery message. 

If you think this can be derived from the new message then sure - but not if it can't.

Upshot, feel free to merge.

Katawann:
@julianoes I removed the energy consumed as suggested by @hamishwillee and I matched order and name based on what have been done for the [battery status v2](https://github.com/mavlink/mavlink/pull/1846/files#diff-413c1cf6a533671c65dda0eeaa0280a129cbf251a06d3ad7ef9ea2bbdb34f1f6R417-R424). Could check it and merged if it's okay ?

protos/telemetry/telemetry.proto
```diff
@@ -575,9 +575,12 @@ enum FixType {
 
 // Battery type.
 message Battery {
-    uint32 id = 3 [(mavsdk.options.default_value)="0"]; // Battery ID, for systems with multiple batteries
-    float voltage_v = 1 [(mavsdk.options.default_value)="NaN"]; // Voltage in volts
-    float remaining_percent = 2 [(mavsdk.options.default_value)="NaN"]; // Estimated battery remaining (range: 0.0 to 1.0)
+    uint32 id = 1 [(mavsdk.options.default_value)="0"]; // Battery ID, for systems with multiple batteries
+    float temperature_cdegc = 2 [(mavsdk.options.default_value)="NaN"]; // Temperature of the battery in centidegrees Celsius. NAN for unknown temperature
```
julianoes, line 579: Why centidegrees? Sorry to bring units up again but that's a bit weird isn't it? When it's a uint type, then that sort of thing makes sense but not for a float, in my opinion.

Katawann: Agree. It is to match what have been done with the [battery v2](https://github.com/mavlink/mavlink/pull/1846/files#diff-413c1cf6a533671c65dda0eeaa0280a129cbf251a06d3ad7ef9ea2bbdb34f1f6R417-R424) but if we agree to have a mismatch I also prefer to go with degree

julianoes: Ok, I'd say please do.

Katawann: Done and ready to be merged :)

julianoes (approved):

Katawann:
Hello team MAVSDK

Any plan to merge this ? Or what is the release policy for MAVSKD proto ?

julianoes:
@Katawann have you already done the MAVSDK part? If so, you can push it and make a PR. That way we know it's ready and once this is merged, the MAVSDK PR will follow right after.

Katawann:
@julianoes work in progress [here](https://github.com/mavlink/MAVSDK/pull/1939)

Sorry for the delay, I needed some time to have clean version of MAVSDK, etc

julianoes (dismissed):

JonasVautherin (approved):

PR #301, opened 2022-11-14
info: add protocol capabilities to the Info plugin

TSC21:
We were missing a way to obtain the autopilot protocol capabilities on the Info plugin,

### DIFF
--- a/protos/info/info.proto
+++ b/protos/info/info.proto
@@ -13,6 +13,8 @@ service InfoService {
     rpc GetFlightInformation(GetFlightInformationRequest) returns(GetFlightInformationResponse) { option (mavsdk.options.async_type) = SYNC; }
     // Get the identification of the system.
     rpc GetIdentification(GetIdentificationRequest) returns(GetIdentificationResponse) { option (mavsdk.options.async_type) = SYNC; }
+    // Get the capabilities of the system.
+    rpc GetCapabilities(GetCapabilitiesRequest) returns(GetCapabilitiesResponse) { option (mavsdk.options.async_type) = SYNC; }
     // Get product information of the system.
     rpc GetProduct(GetProductRequest) returns(GetProductResponse) { option (mavsdk.options.async_type) = SYNC; }
     // Get the version information of the system.
@@ -31,6 +33,12 @@ message GetIdentificationRequest {}
 message GetIdentificationResponse {
     InfoResult info_result = 1;
     Identification identification = 2; // Identification of the system
+}
+
+message GetCapabilitiesRequest {}
+message GetCapabilitiesResponse {
+    InfoResult info_result = 1;
+    Capabilities capabilities = 2; // Capabilities of the system
 }
 
 message GetProductRequest {}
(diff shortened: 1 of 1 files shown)


JonasVautherin (commented):
This is exposing MAVLink at a pretty low level, isn't it?

Isn't this something that should be handled in the MAVSDK implementation, instead of exposed in the MAVSDK API?

@julianoes do you have an opinion?

TSC21:
> Isn't this something that should be handled in the MAVSDK implementation, instead of exposed in the MAVSDK API?

I don't disagree, but how would you expose it higher level then? Process the bitmap entries at the implementation level and then expose the resulting capabilities as boolean flags or something? I just went through the simplest process here.

JonasVautherin:
For instance the `mission` plugin is supposed to take care of the mission protocol, so the user should not have to do anything with `PROTOCOL_CAPABILITY_MISSION_FLOAT` or `PROTOCOL_CAPABILITY_MISSION_INT` (actually what would they do with this info?).

If I take e.g. `PROTOCOL_CAPABILITY_FLIGHT_TERMINATION`, then I guess that `action->terminate()` could check that flag and return an `UNSUPPORTED` result (right now probably it just times out).

So maybe I can ask this: which flag are you planning to use, and how?

TSC21:
> For instance the `mission` plugin is supposed to take care of the mission protocol, so the user should not have to do anything with `PROTOCOL_CAPABILITY_MISSION_FLOAT` or `PROTOCOL_CAPABILITY_MISSION_INT` (actually what would they do with this info?).
> 
> If I take e.g. `PROTOCOL_CAPABILITY_FLIGHT_TERMINATION`, then I guess that `action->terminate()` could check that flag and return an `UNSUPPORTED` result (right now probably it just times out).
> 
> So maybe I can ask this: which flag are you planning to use, and how?

Right now, the the MAVLink test suite, I have to check the following capabilities:

- `MAV_PROTOCOL_CAPABILITY_COMMAND_INT`
- `MAV_PROTOCOL_CAPABILITY_MISSION_FENCE`
- `MAV_PROTOCOL_CAPABILITY_MISSION_RALLY`

And might have to expand to others.

JonasVautherin:
> it purely uses passthough

And can't you access those capabilities from passthrough?

TSC21:
> > it purely uses passthough
> 
> And can't you access those capabilities from passthrough?

Technically I could, but there's a conceptual difference in the usage: on the test suite, passthrough is used to validate if a certain message is being received at a certain rate (this means that the autopilot is sending that data, and we validate it). The passthrough is not used to fetch data to setup the tests, is used during the tests themselves. We before had these capabilities exposed through the `get_autopilot_version_data()` function, and that function was part of the mavsdk::system, and now we don't have it anymore.

JonasVautherin:
@julianoes what's your opinion there?

JonasVautherin:
@TSC21 for the record, didn't you say that you could access that fairly easily from mavlink-passthrough?

protos/info/info.proto
```diff
@@ -63,6 +71,34 @@ message Identification {
...
+        PROTOCOL_CAPABILITY_SET_ACTUATOR_TARGET	= 0x00000400; // Autopilot supports direct actuator control.
+        PROTOCOL_CAPABILITY_FLIGHT_TERMINATION = 0x00000800; // Autopilot supports the MAV_CMD_DO_FLIGHTTERMINATION command (flight termination).
+        PROTOCOL_CAPABILITY_COMPASS_CALIBRATION = 0x00001000; // Autopilot supports onboard compass calibration.
+        PROTOCOL_CAPABILITY_MAVLINK2 = 0x00002000; // Autopilot supports MAVLink version 2.
+        PROTOCOL_CAPABILITY_MISSION_FENCE = 0x00004000; // Autopilot supports mission fence protocol.
+        PROTOCOL_CAPABILITY_MISSION_RALLY = 0x00008000; // Autopilot supports mission rally point protocol.
+        PROTOCOL_CAPABILITY_RESERVED2 = 0x00010000; // Reserved for future use.
+        PROTOCOL_CAPABILITY_PARAM_ENCODE_C_CAST = 0x00020000; // Parameter protocol uses C-cast of parameter values to set the param_value (float) fields: https://mavlink.io/en/services/parameter.html#parameter-encoding. Note that either this flag or PROTOCOL_CAPABILITY_PARAM_ENCODE_BYTEWISE should be set if the parameter protocol is supported.
```
julianoes, line 96: Pretty sure this does not work with the auto-generation. The only way to make this work is to change how the auto-generation works, or add boolean flags for each of the flags.

julianoes (changes requested):
I don't think it makes sense to expose this. It's as @JonasVautherin said "more of a MAVLink internal thing", exactly the sort of thing that a user of MAVLink should not have to think about.

I think that testing that the internals of MAVLink are working correctly is not a primary use case of MAVSDK, and so I would probably use MAVLink passthrough for that.

julianoes:
Closing as this does not work and it's not the intent to leak these MAVLink internal flags in the API. It would make more sense to use MAVLinkPassthrough for this.

PR #303, opened 2022-11-24
geofence: Add support for circular geofences

YalamareddyChaitanya:
It would be really nice to have support for circular geofences in Geofence plugin.

### DIFF
--- a/protos/geofence/geofence.proto
+++ b/protos/geofence/geofence.proto
@@ -2,15 +2,17 @@ syntax = "proto3";
 
 package mavsdk.rpc.geofence;
 
+import "mavsdk_options.proto";
+
 option java_package = "io.mavsdk.geofence";
 option java_outer_classname = "GeofenceProto";
 
 // Enable setting a geofence.
 service GeofenceService {
     /*
-     * Upload a geofence.
+     * Upload geofences.
      *
-     * Polygons are uploaded to a drone. Once uploaded, the geofence will remain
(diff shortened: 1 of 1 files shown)


protos/geofence/geofence.proto
```diff
@@ -26,20 +28,33 @@ message Point {
...
+message Circle {
+    Point point = 1; // Point defining the center
+    float radius = 2 [(mavsdk.options.default_value)="NaN"]; // Radius of the circular fence
+    FenceType fence_type = 3; // Fence type
+}
+
+// Geofence data type.
+message GeofenceData {
     repeated Polygon polygons = 1; // Polygon(s) representing the geofence(s)
+    repeated Circle circles = 2; // Circle(s) representing the geofence(s)
```
julianoes, line 53: @JonasVautherin that sort of extension should be ok and backwards compatible, right?

I guess anyway, this will go into MAVSDK 2.0.

julianoes (approved):
Nice, thanks for the PR!

protos/geofence/geofence.proto
JonasVautherin: At the protobuf level, yes. But when autogenerating our API, I think it will break it (now the constructor will take two objects instead of one).

JonasVautherin (approved):

protos/geofence/geofence.proto
julianoes: I think that's ok for 2.0 then.

PR #304, opened 2022-12-12
log_files: remove synchronous DownloadLogFile

dayjaby:
fixes https://github.com/mavlink/MAVSDK-Python/issues/534

### DIFF
--- a/protos/log_files/log_files.proto
+++ b/protos/log_files/log_files.proto
@@ -17,10 +17,7 @@ service LogFilesService {
         option (mavsdk.options.async_type) = ASYNC;
         option (mavsdk.options.is_finite) = true;
     }
-    // Download log file synchronously.
-    rpc DownloadLogFile(DownloadLogFileRequest) returns(DownloadLogFileResponse) {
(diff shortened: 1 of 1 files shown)


JonasVautherin (approved):

julianoes (approved):
Thanks!

julianoes:
I don't think that's the right fix. We will have the same problem in other places, right? Or why would this one be special?

JonasVautherin:
This is the only place where we have both `SubscribeDownloadFile` and `DownloadFile`. That should be forbidden, because both end up generating a function called `download_file()` in Python and `downloadFile` in Java. The `Subscribe-` prefix just indicates that it is a stream.

Arguably a better fix would be to detect that it is a stream with the `stream` keyword in the response, and keep `Subscribe` in the name. But well... that's how it is now :sweat_smile:

julianoes:
Oh, I see. That should be forbidden indeed.

julianoes:
@JonasVautherin so I guess we can leave this as is and go through with the changes in C++, towards v2.0? Right?

JonasVautherin:
Yes!

PR #308, opened 2023-01-20
Winch and Gripper plugins

potaito:
## Related PRs
- Proto: https://github.com/mavlink/MAVSDK-Proto/pull/308 **(this PR)**
- MAVSDK: https://github.com/mavlink/MAVSDK/pull/1958
- MAVSDK-Python: MAVSDK-Python: https://github.com/mavlink/MAVSDK-Python/pull/545 

## Description
Message definitions for Winches and Grippers. This reflects the following mavlink messages:
- MAV_CMD_DO_GRIPPER [(211)](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GRIPPER)
- MAV_CMD_DO_WINCH ([42600](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_WINCH) )

and their submessages for the gripper and winch actions, as well as the winch status.

Please don't merge before I squashed the commits

### DIFF
--- a/protos/gripper/gripper.proto
+++ b/protos/gripper/gripper.proto
@@ -18,13 +18,14 @@ service GripperService {
 }
 
 enum GripperAction {
-    GRIPPER_ACTION_RELEASE = 0;
-    GRIPPER_ACTION_GRAB = 1;
+    GRIPPER_ACTION_RELEASE = 0; // Open the gripper to release the cargo
+    GRIPPER_ACTION_GRAB = 1; // Close the gripper and grab onto cargo
 }
 
 message GrabRequest {
     uint32 instance = 1;
 }
+
 message GrabResponse {
     GripperResult gripper_result = 1;
 }
@@ -32,6 +33,7 @@ message GrabResponse {
 message ReleaseRequest {
     uint32 instance = 1;
 }
+
 message ReleaseResponse {
     GripperResult gripper_result = 1;
 }
@@ -40,13 +42,13 @@ message ReleaseResponse {
 message GripperResult {
     // Possible results returned for gripper action requests.
     enum Result {
-        RESULT_UNKNOWN = 0; // Unknown result
-        RESULT_SUCCESS = 1; // Request was successful
-        RESULT_NO_SYSTEM = 2; // No system is connected
-        RESULT_BUSY = 3;    // Temporarily rejected
-        RESULT_TIMEOUT = 4; // Request timed out
+        RESULT_UNKNOWN = 0;     // Unknown result
+        RESULT_SUCCESS = 1;     // Request was successful
+        RESULT_NO_SYSTEM = 2;   // No system is connected
+        RESULT_BUSY = 3;        // Temporarily rejected
+        RESULT_TIMEOUT = 4;     // Request timed out
         RESULT_UNSUPPORTED = 5; // Action not supported
-        RESULT_FAILED = 6; // Action failed
+        RESULT_FAILED = 6;      // Action failed
     }
 
     Result result = 1; // Result enum value
--- a/protos/gripper/gripper.proto
+++ b/protos/gripper/gripper.proto
@@ -1,0 +1,54 @@
+syntax = "proto3";
+
+package mavsdk.rpc.gripper;
+
+option java_package = "io.mavsdk.gripper";
+option java_outer_classname = "GripperProto";
+
+service GripperService {
+    /*
+     * Gripper grab cargo.
+     */
+    rpc Grab(GrabRequest) returns(GrabResponse) {}
+
+    /*
+     * Gripper release cargo.
+     */
+    rpc Release(ReleaseRequest) returns(ReleaseResponse) {}
+}
+
+enum GripperAction {
+    GRIPPER_ACTION_RELEASE = 0;
+    GRIPPER_ACTION_GRAB = 1;
+}
+
+message GrabRequest {
+    uint32 instance = 1;
+}
+message GrabResponse {
+    GripperResult gripper_result = 1;
+}
+
+message ReleaseRequest {
+    uint32 instance = 1;
+}
+message ReleaseResponse {
+    GripperResult gripper_result = 1;
+}
+
+// Result type.
+message GripperResult {
+    // Possible results returned for gripper action requests.
+    enum Result {
+        RESULT_UNKNOWN = 0; // Unknown result
+        RESULT_SUCCESS = 1; // Request was successful
+        RESULT_NO_SYSTEM = 2; // No system is connected
+        RESULT_BUSY = 3;    // Temporarily rejected
+        RESULT_TIMEOUT = 4; // Request timed out
+        RESULT_UNSUPPORTED = 5; // Action not supported
+        RESULT_FAILED = 6; // Action failed
+    }
+
+    Result result = 1; // Result enum value
+    string result_str = 2; // Human-readable English string describing the result
+}
--- a/protos/winch/winch.proto
+++ b/protos/winch/winch.proto
@@ -120,7 +120,7 @@ message RelaxResponse {
 message RelativeLengthControlRequest {
     uint32 instance = 1; // Instance ID of the winch addressed by this request
     float length_m = 2; // Length of line to unwind or wind
-    float rate_m_s = 3 [(mavsdk.options.default_value)="NaN"];; // Rate at which to wind or unwind the line
+    float rate_m_s = 3 [(mavsdk.options.default_value)="NaN"]; // Rate at which to wind or unwind the line
 }
 
 message RelativeLengthControlResponse {
--- a/protos/winch/winch.proto
+++ b/protos/winch/winch.proto
@@ -1,0 +1,200 @@
+syntax = "proto3";
+
+package mavsdk.rpc.winch;
+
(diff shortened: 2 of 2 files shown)


protos/gripper/gripper.proto
```diff
@@ -0,0 +1,61 @@
...
+    // Possible results returned for action requests.
+    enum Result {
+        RESULT_UNKNOWN = 0; // Unknown result
+        RESULT_SUCCESS = 1; // Request was successful
+        RESULT_NO_SYSTEM = 2; // No system is connected
+        RESULT_CONNECTION_ERROR = 3; // Connection error
+        RESULT_BUSY = 4; // Vehicle is busy
+        RESULT_COMMAND_DENIED = 5; // Command refused by vehicle
+        RESULT_COMMAND_DENIED_LANDED_STATE_UNKNOWN = 6; // Command refused because landed state is unknown
+        RESULT_COMMAND_DENIED_NOT_LANDED = 7; // Command refused because vehicle not landed
```
JonasVautherin, line 50: Do those make sense for a gripper? :thinking:

protos/gripper/gripper.proto
```diff
@@ -0,0 +1,61 @@
...
+        RESULT_TIMEOUT = 8; // Request timed out
+        RESULT_VTOL_TRANSITION_SUPPORT_UNKNOWN = 9; // Hybrid/VTOL transition support is unknown
+        RESULT_NO_VTOL_TRANSITION_SUPPORT = 10; // Vehicle does not support hybrid/VTOL transitions
+        RESULT_PARAMETER_ERROR = 11; // Error getting or setting parameter
+        RESULT_UNSUPPORTED = 12; // Action not supported
```
JonasVautherin, line 55: Same here: does that happen for the gripper?

protos/winch/winch.proto
```diff
@@ -0,0 +1,207 @@
...
+
+message SubscribeStatusRequest {}
+message StatusResponse {
+    Status status = 1; // The next 'winch status' state
+}
+
+message StatusFlags {
+    bool healthy = 1;           // Winch is healthy
+    bool fully_retracted = 2;   // Winch line is fully retracted
+    bool moving = 3;            //Winch motor is moving
```
JonasVautherin, line 74:
```suggestion
    bool moving = 3;            // Winch motor is moving
```

protos/winch/winch.proto
```diff
@@ -0,0 +1,207 @@
...
+        RESULT_CONNECTION_ERROR = 3; // Connection error
+        RESULT_BUSY = 4; // Vehicle is busy
+        RESULT_COMMAND_DENIED = 5; // Command refused by vehicle
+        RESULT_COMMAND_DENIED_LANDED_STATE_UNKNOWN = 6; // Command refused because landed state is unknown
+        RESULT_COMMAND_DENIED_NOT_LANDED = 7; // Command refused because vehicle not landed
+        RESULT_TIMEOUT = 8; // Request timed out
+        RESULT_VTOL_TRANSITION_SUPPORT_UNKNOWN = 9; // Hybrid/VTOL transition support is unknown
+        RESULT_NO_VTOL_TRANSITION_SUPPORT = 10; // Vehicle does not support hybrid/VTOL transitions
+        RESULT_PARAMETER_ERROR = 11; // Error getting or setting parameter
+        RESULT_UNSUPPORTED = 12; // Action not supported
```
JonasVautherin, line 201: Same question here, I wonder if those results make sense for a winch

JonasVautherin (changes requested):
Looking nice! :rocket:

protos/gripper/gripper.proto
julianoes: I think it could if the gripper commands are not implemented, e.g. for an older PX4 version.

protos/gripper/gripper.proto
potaito: Ah yes, sorry. I forgot to narrow these down. I copied the struct from `actions.proto` initially :see_no_evil: 

I have to check which responses are possible to a vehicle command.

protos/gripper/gripper.proto
potaito: `RESULT_UNSUPPORTED` is one of the few that makes sense

potaito: Would it make sense to support _all_ of the possible results that the command sender can return?

https://github.com/mavlink/MAVSDK/blob/444d6b693b2d76cd594b1df074f28f468cc78f47/src/mavsdk/core/mavlink_command_sender.h#L24-L37

In case of simple systems like winches and grippers, the results `NoSystem`, `ConnectionError`, and `Timeout` for example are basically the same.

potaito: Aha, it seems that the grpc code auto-generation at least expects there to be a `NoSystem` result, so I can't remove that one.

JonasVautherin (approved):
:rocket:

potaito:
@JonasVautherin @julianoes there are now the following possible results:
- `RESULT_UNKNOWN`, as a default fallthrough should MAVSDK receive something it cannot handle
- `RESULT_SUCCESS`, obvious
- `RESULT_NO_SYSTEM` is required by MAVSDK
- `RESULT_BUSY` corresponds to mavlink's [MAV_RESULT_TEMPORARILY_REJECTED](https://mavlink.io/en/messages/common.html#MAV_RESULT_TEMPORARILY_REJECTED) as far as I can tell. A winch or gripper _might_ send this while it's already performing an operation. These can take 30 seconds or more.
- ~`RESULT_COMMAND_DENIED` for supported commands with invalid parameters~. For winches and grippers this is probably not that important, since there are only few parameters to begin with, and the winch mavlink message for example states, that parameters are simply ignored. I think I'm going to remove this result as well...
- `RESULT_TIMEOUT` for when the winch or gripper is not responding to commands with an ACK.
- `RESULT_UNSUPPORTED` makes sense to have. Particularly for winches, because there are currently 10 winch actions, and no winch I know off supports _all_ of these, but only a subset of actions.
- `RESULT_FAILED` for when an action was acknowledged and then attempted to be executed, but something went wrong. For example, the over-current protection kicked in because the winch's rope is caught somewhere.

protos/gripper/gripper.proto
julianoes: They are different though.

potaito:
Just squashed the whole thing into a single commit

JonasVautherin (dismissed):
:rocket:

protos/winch/winch.proto
```diff
@@ -0,0 +1,200 @@
...
+    WINCH_ACTION_RELAXED = 0;
+    WINCH_ACTION_RELATIVE_LENGTH_CONTROL = 1;
+    WINCH_ACTION_RATE_CONTROL = 2;
+    WINCH_ACTION_LOCK = 3;
+    WINCH_ACTION_DELIVER = 4;
+    WINCH_ACTION_HOLD = 5;
+    WINCH_ACTION_RETRACT = 6;
+    WINCH_ACTION_LOAD_LINE = 7;
+    WINCH_ACTION_ABANDON_LINE = 8;
+    WINCH_ACTION_LOAD_PAYLOAD = 9;
```
julianoes, line 109: Probably requires docstrings.

protos/winch/winch.proto
```diff
@@ -0,0 +1,200 @@
...
+    uint32 instance = 1;
+}
+message RelaxResponse {
+    WinchResult winch_result = 1;
+}
+
+message RelativeLengthControlRequest {
+    uint32 instance = 1;
+    float length = 2;
+    float rate = 3;
```
julianoes, line 122: Could you add units? Something like:

```suggestion
    float length_m = 2;
    float rate_hz = 3;
```

julianoes (commented):
Nice, thanks for that!

protos/winch/winch.proto
potaito: Ah yes, of course :+1:

protos/winch/winch.proto
```diff
@@ -0,0 +1,210 @@
...
+
+message Status {
+    uint64 time_usec = 1;
+    float line_length_m = 2;
+    float speed_m_s = 3;
+    float tension_kg = 4;
+    float voltage_v = 5;
+    float current_a = 6;
+    int32 temperature_c = 7;
+    StatusFlags status_flags = 8;
```
JonasVautherin, line 96: We need those as well otherwise `protoc-gen-mavsdk` will generate weird emptry docstrings :see_no_evil: 

```suggestion
    uint64 time_usec = 1; // Time in usec
    float line_length_m = 2; // Length of the line in meters
    float speed_m_s = 3; // Speed in meters per second
    float tension_kg = 4; // Tension in kilograms
    float voltage_v = 5; // Voltage in volts
    float current_a = 6; // Current in amperes
    int32 temperature_c = 7; // Temperature in Celsius
    StatusFlags status_flags = 8; // Status flags
```

potaito: Thanks!!

JonasVautherin (dismissed):

JonasVautherin (approved):

potaito:
Sorry, that's what happens when I use the github web GUI to make changes to a file :sweat_smile: 
CI is now fixed.

PR #312, opened 2023-03-09
add SetRateSysStatus

41Mo:
# Related PRs

- MAVSDK: https://github.com/mavlink/MAVSDK/pull/1998

# Description:

This pr adds a new SetRateSysStatus request that will improve API usage as described in mavlink/MAVSDK#1996.

### DIFF
--- a/protos/telemetry/telemetry.proto
+++ b/protos/telemetry/telemetry.proto
@@ -129,6 +129,8 @@ service TelemetryService {
     rpc SetRateDistanceSensor(SetRateDistanceSensorRequest) returns(SetRateDistanceSensorResponse) {}
     // Set rate to 'Altitude' updates.
     rpc SetRateAltitude(SetRateAltitudeRequest) returns(SetRateAltitudeResponse){}
+    // Set rate to 'SYS_STATUS' msg updates, which allows to recieve vehicle HEALTH information.
+    rpc SetRateSysStatus(SetRateSysStatusRequest) returns(SetRateSysStatusResponse) {}
 
     // Get the GPS location of where the estimator has been initialized.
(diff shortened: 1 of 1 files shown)


julianoes (commented):
I would call it SetRateHealth, otherwise it's not clear what it changes.
What do you think?

41Mo:
SYS_STATUS message is a little confusing, because it contains not only health data, but also basic battery information.
So i am not sure, how to properly call it. So what do you think?

41Mo:
SYS_STATUS message is a little confusing, because it contains not only health data, but also basic battery information.
I am not sure how to name it properly. What do you think, just HEALTH is enough?

41Mo:
In my opinion better call it SYS_STATUS and give a reference to mavlink message defenition and small commentary about content of this message

julianoes:
Right, I can see how it's confusing. The other question is if PX4 actually allows you to send the sys status quicker.

Or what about just not allowing this message to be sent at a higher rate? Is it really required at a higher rate?

41Mo:
@julianoes  yes, good point. In my practice, there was no case when it was required to receive the status with greater frequency.

Maybe then just add the frequency setting without additional api, directly to the plugin's init?

julianoes:
Ok, and you would request it at 1Hz?

41Mo:
Yes

julianoes:
Ok makes sense. That's for the cases where it's not already automatically sent at 1Hz.

julianoes:
I'm sorry I didn't act on this earlier. Sometimes life takes over.

Anyway, it looks like the battery rate has been fixed by https://github.com/mavlink/MAVSDK/pull/2021 and I doubt we need health information faster than 1 Hz. I believe PX4 won't send it out quicker anyway as the pre-flight checks don't run faster than 1 Hz anyway.

 If I'm wrong with that let me know.

PR #313, opened 2023-03-27
[DRAFT] Landing Target Plugin

potaito:
This is far from ready to be merged, but I wanted to show my intent to implement a plugin for precision landing by opening this PR.

This is eventually going to include the new precision target messages being proposed for mavlink: https://github.com/mavlink/mavlink/pull/1943

I don't know if we should bother to initially support the original LANDING_TARGET mavlink message, which the aforementioned PR is going to replace: https://mavlink.io/en/messages/common.html#LANDING_TARGET

Notes for the API itself:
- The measurements are just that: Measurements, or sensor observations. They should not dictate the task that should be executed. As such the API should not mention LANDING_TARGET or anything else landing related.
- The API should be usable to publish absolute (lat/lon) and relative data (relative to drone, sensor, or NED frame of drone).

### DIFF
--- a/protos/landing_target/landing_target.proto
+++ b/protos/landing_target/landing_target.proto
@@ -1,18 +1,18 @@
 syntax = "proto3";
 
-package mavsdk.rpc.landing_target_server;
+package mavsdk.rpc.landing_target;
 
 import "mavsdk_options.proto";
 
-option java_package = "io.mavsdk.landing_target_server";
-option java_outer_classname = "LandingTargetServerProto";
+option java_package = "io.mavsdk.landing_target";
+option java_outer_classname = "LandingTargetProto";
 
 /*
  * Allows developers to stream the position of a landing target for the drone to land on.
  */
-service LandingTargetServerService {
+service LandingTargetService {
     // Publish landing target measurement recorded by a sensor onboard the drone
-    rpc PublishLandingTargetRelative(PublishLandingTargetRelativeRequest) returns(PublishLandingTargetRelativeResponse) { option (mavsdk.options.async_type) = SYNC; }
+    rpc PublishPositionRelative(PublishPositionRelativeRequest) returns(PublishPositionRelativeResponse) { option (mavsdk.options.async_type) = SYNC; }
 }
 
 // Position type in sensor frame coordinates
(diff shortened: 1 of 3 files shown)


JonasVautherin:
> I don't know if we should bother to initially support the original LANDING_TARGET

If MAVSDK does not support it yet and nobody ever asked for it, let's just ignore it :blush:.

potaito:
@JonasVautherin Makes sense, that was my feeling too. The new mavlink message is about to be merged into the development dialect. Is the develop dialect okay to be used in MAVSDK?  I don't really see any reason against it...

potaito:
@JonasVautherin Makes sense, that was my feeling too. The new mavlink message is about to be merged into the development dialect. Is the develop dialect okay to be used in MAVSDK?  I don't really see any reason against it...

JonasVautherin:
Right now we are on [`common`](https://github.com/mavlink/MAVSDK/blob/main/third_party/mavlink/CMakeLists.txt#L12). Not sure about `development`.

I guess the problem is that `development` could be fundamentally changed (breaking the MAVSDK API, forcing a new major version), and I don't know how compatible that is with MAVSDK being a "stable" API (because `development` is unstable by definition, right?).

@julianoes any opinion here?

JonasVautherin:
Right now we are on [`common`](https://github.com/mavlink/MAVSDK/blob/main/third_party/mavlink/CMakeLists.txt#L12). Not sure about `development`.

I guess the problem is that `development` could be fundamentally changed (breaking the MAVSDK API, forcing a new major version), and I don't know how compatible that is with MAVSDK being a "stable" API (because `development` is unstable by definition, right?).

@julianoes any opinion here?

potaito:
Yep, that's why I ask :smile: 

Currently the only supported way to send data for precision landing with the common dialect is with the old, borked and funky LANDING_TARGET message in the common dialect. 

What I can do is implement a basic proto API that covers only the features that are offered by the old and the new mavlink message. Then we can swap out the mavlink message in the background, and afterwards it can be extended by new capabilities only offered by the new messages.

JonasVautherin:
But wouldn't that still require MAVSDK to compile all the `development.xml` messages?

julianoes:
As soon as we have it working end to end (maybe with a video in the MAVLink call) then we should be able to move it into common and then be able to merge it here.

Alternatively, we could have some sort of development setup in MAVSDK that allows to use the development.xml dialect. The main issue is that we don't want to release anything depending on development.

potaito:
Hey @julianoes, thanks for your input.
I think in this case the corresponding mavlink message should stay in development for a bit so that multiple parties get a chance to try it out and we can make sure that it covers 99% of the use cases.

This means I'll just implement the common feature set for now. Think that's the easiest approach.

julianoes:
@potaito :wave: what's the status here? Do you know whether this is in PX4 and used yet?

potaito:
@julianoes the draft here is matching PX4's current implementation / capabilities. I had a need for this feature in MAVSKD a couple of times over the last years. But currently I don't :sweat_smile: So I'm not sure when / if I'll get around to doing the implementation.

PR #317, opened 2023-06-11
param_server: allow to change parameter internally

dayjaby:
Allow a param_server to change a parameter that it is providing.


JonasVautherin (commented):
Can't we (couldn't we?) do that by calling `ProvideParam` again with the same name but a different value?

julianoes:
I think this makes sense. The tricky thing is that you're supposed to "provide" all params on init, before any mavlink traffic happens, and the number of params can't change later.

JonasVautherin:
> The tricky thing is that you're supposed to "provide" all params on init, before any mavlink traffic happens

Oh so that can't work with the language bindings, right?

dayjaby:
Language bindings should be fine. So a python mavsdk app for example can represent an entire mavlink component, using the param_server plugin to provide all the parameters. If something happens within the component, the same plugin should be able to be used to "change" one of these parameters.

Issue #316, opened 2023-05-23
Design camera server implement discussion


tbago:
When check the **camera.proto** and **camera_server.proto**,  I found something need a discussion.

1. The camera.proto and camera_server.proto have some same definition (e.g. [Position in client](https://github.com/mavlink/MAVSDK-Proto/blob/82e8540f980ea318447f79f4cba8813993ae2a6b/protos/camera_server/camera_server.proto#L79) [Position in server](https://github.com/mavlink/MAVSDK-Proto/blob/82e8540f980ea318447f79f4cba8813993ae2a6b/protos/camera_server/camera_server.proto#L79) ). Maybe better define in the camera_common.proto. The camera_server.proto also need to add [camera_mode](https://github.com/mavlink/MAVSDK-Proto/blob/82e8540f980ea318447f79f4cba8813993ae2a6b/protos/camera/camera.proto#L255) [camera_status](https://github.com/mavlink/MAVSDK-Proto/blob/82e8540f980ea318447f79f4cba8813993ae2a6b/protos/camera/camera.proto#L349)

2. The [camera status](https://github.com/mavlink/MAVSDK-Proto/blob/82e8540f980ea318447f79f4cba8813993ae2a6b/protos/camera/camera.proto#L349) proto define. The camera status is packed in two mavlink message(**MAVLINK_MSG_ID_CAMERA_CAPTURE_STATUS** **MAVLINK_MSG_ID_STORAGE_INFORMATION**). The camera need to check the two [status value](https://github.com/mavlink/MAVSDK/blob/main/src/mavsdk/plugins/camera/camera_impl.cpp#L1367). When implement the camera server side also have the same problem.

3. The camera parameter is not in the param list and need to be set by [**ParamServer**](https://github.com/mavlink/MAVSDK-Proto/blob/82e8540f980ea318447f79f4cba8813993ae2a6b/protos/param_server/param_server.proto#LL47C9-L47C27). 
4. But the value type is just [string type](https://github.com/mavlink/MAVSDK-Proto/blob/82e8540f980ea318447f79f4cba8813993ae2a6b/protos/param_server/param_server.proto#L98). I read some [camera define xml](https://github.com/mavlink/MAVSDK/blob/main/src/mavsdk/plugins/camera/camera_definition_files/e90.xml#L87).The define type has various types. So I want to change the param_server.proto to support more custom define type.

```
enum ParamType {
    INT16_TYPE = 1;
    INT32_TYPE = 2;
[...]

PR #320, opened 2023-06-17
Add more camera function

tbago:
add start stop video streaming in camera server, add streaming id in camera side (the camera can support multi video stream, so the stream id is useful).


tbago:
In the pre [issue](https://github.com/mavlink/MAVSDK-Proto/issues/316), we have alreay have some discuss.

1、For now I think the duplication of proto is ok too;
2、In the camera server side, I just implement the **MAVLINK_MSG_ID_CAMERA_CAPTURE_STATUS** **MAVLINK_MSG_ID_STORAGE_INFORMATION** in separeate message.  two separate message is easy to implement but is not the same in the [camera](https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/camera/camera.proto#L349) side.

tbago:
If the following commit is ok, then I will continue to add the following funcitons:
1、reset_settings and format storage;

2、set video streaming info in camera server to make the **[subscirbe video streaming](https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/camera/camera.proto#LL71C6-L71C6)** working, (support multi video streaming);

3、Add support manual set [the camera definition file ](https://github.com/mavlink/MAVSDK/blob/main/src/mavsdk/plugins/camera/camera_definition_files/e50.xml) outside.  I know in the camera already support[ http download ](https://github.com/mavlink/MAVSDK/blob/main/src/mavsdk/plugins/camera/camera_impl.cpp#LL1220C18-L1220C42)inside.
But if I want to support use mavlink ftp to download the definition file, I need to tell the **MAVSDK** the download file location. So I prefer to download the file in client side, and just pass the definition file data to **MAVSDK**.  I think use the http download inside the **MAVSDK** will make it easy to use. So  one more suggestion, if the mavlink ftp can support download the file in memory maybe more better.
**Sorry for my poor english**,  my suggestion is there are two way to implement the set camera definition file(support both http and mavlink ftp)
1、best way, the mavlink ftp can support download file in memory, and the **MAVSDK** user will not care about the download method;
2、better way, support manual set camera definition data, and the **MAVSDK** user can use anyway they like to download the definition file.

protos/camera_server/camera_server.proto
```diff
@@ -143,3 +203,63 @@ message CameraServerResult {
...
+        IMAGE_STATUS_INTERVAL_IDLE = 2; // interval set but idle
+        IMAGE_STATUS_INTERVAL_IN_PROGRESS = 3; // interval set and capture in progress)
+    }
+    enum VideoStatus {
+        VIDEO_STATUS_IDLE = 0;  // idle
+        VIDEO_STATUS_CAPTURE_IN_PROGRESS = 1;   // capture in progress
+    }
+    float image_interval = 1; // Image capture interval (in s)
+    float recording_time_s = 2; // Elapsed time since recording started (in s)
+    float available_capacity = 3; // Available storage capacity. (in MiB)
```
julianoes, line 261: I suggest to add the units like the other fields:

```suggestion
    float available_capacity_mib = 3; // Available storage capacity. (in MiB)
```

And in other places too.

julianoes (commented):
Thanks, see comment regarding units.

JonasVautherin (dismissed):

protos/camera_server/camera_server.proto
```diff
@@ -143,3 +203,63 @@ message CameraServerResult {
...
+    float total_storage_mib = 3; // Total storage (in MiB)
+
+    StorageStatus storage_status = 4; // Storage status
+
+    uint32 storage_id = 5; // Storage ID starting at 1
+
+    StorageType storage_type = 6; // Storage type
+
+    float read_speed = 7; // Read speed [MiB/s]
+    float write_speed = 8; // Write speed [MiB/s]
```
julianoes, line 245: Also here:
```suggestion
    float read_speed_mib_s = 7; // Read speed [MiB/s]
    float write_speed_mib_s = 8; // Write speed [MiB/s]
```

tbago:
This PR is all about the basic camera function. 
For more function, I will create new PR. So for now the PR is complete.

protos/camera_server/camera_server.proto
```diff
@@ -58,6 +85,55 @@ message StopVideoResponse {
     int32 stream_id = 1;    // video stream id
 }
 
+message SubscribeStartVideoStreamingRequest {}
+message StartVideoStreamingResponse {
+    int32 stream_id = 1;    // video stream id
+}
```
julianoes, line 91: Shouldn't this also include a result? Otherwise, what if the video stream starting failed?

protos/camera_server/camera_server.proto
```diff
@@ -58,6 +85,55 @@ message StopVideoResponse {
...
+
+message SubscribeStopVideoStreamingRequest {}
+message StopVideoStreamingResponse {
+    int32 stream_id = 1;    // video stream id
```
julianoes, line 95: Same here, should it have a result?

protos/camera_server/camera_server.proto
```diff
@@ -58,6 +85,55 @@ message StopVideoResponse {
...
+
+message SubscribeSetModeRequest {}
+message SetModeResponse {
+    Mode mode = 1;
+}
+
+message SubscribeStorageInformationRequest {}
+message StorageInformationResponse {
+    int32 storage_id = 1;
+}
```
julianoes, line 106: Shouldn't this contain the storage information?

E.g. if you look at the action server, this is what the response contains:
https://github.com/mavlink/MAVSDK-Proto/blob/8966cf59029ac17d81654071ca6d3d27f0d8897d/protos/action_server/action_server.proto#L79-L82

protos/camera_server/camera_server.proto
```diff
@@ -58,6 +85,55 @@ message StopVideoResponse {
...
+
+message RespondStorageInformationRequest {
+    StorageInformation storage_information = 1; // camera storage information
+}
+message RespondStorageInformationResponse {
+    CameraServerResult camera_server_result = 1;
+}
```
julianoes, line 113: And then these shouldn't be required because the response is sent in the response to the subscription.

Or do I have that completely backwards in my head now @JonasVautherin?

julianoes (changes requested):
I think we need to rethink the API. See inline comment.

protos/camera_server/camera_server.proto
tbago: I have some misunderstood about the camera server design.

For now I add all respond command with feedback.

julianoes: But that's different from how it is done in other server plugins, no?

tbago: Oh, I just use the origin take photo way to implment other camera function. Can you provide another suggestion way to implment the interface.

julianoes: Like this for instance?

https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/action_server/action_server.proto#L74C9-L77

https://github.com/mavlink/MAVSDK/blob/main/src/mavsdk/plugins/action_server/action_server_impl.cpp#L63-L96

julianoes: Does that make sense @tbago ? Do you understand what I mean?

tbago: Hi julian, I try to undersand the action server implement. Sorry for my poor english, I just draw a graph to show what I mean.
``` mermaid
---
title: Action Process
---
flowchart LR
    subgraph MAVSDK Client
    C1[Action]
    end
    
    subgraph MavSDK Server
    M1[Receive Action message]
    M2[Process action]
    M1<-->M2
    end

    subgraph ActionSubscribe
    B1[subscribe result]
    end

    C1 <-->|mavlink| M1
    M2 <--> |subscribe| B1
```
The [subscribe arm](https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/action_server/action_server.proto#L13) is just callback function for subscribe the action server result and arm state. And the action result is process by [this code](https://github.com/mavlink/MAVSDK/blob/main/src/mavsdk/plugins/action_server/action_server_impl.cpp#L70C23-L90).

tbago: For the camera function. I just want to use the [camera server](https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/camera_server/camera_server.proto) like middleware. And the camera process logic will be in the backend, like the [example code](https://github.com/tbago/MAVSDK/blob/add_more_camera_function/examples/camera_server/camera_server.cpp#L111). And I will use [respond message](https://github.com/tbago/MAVSDK-Proto/blob/add_more_camera_function/protos/camera_server/camera_server.proto#L27) to tell the **MAVSDK Camera Server** process result.

tbago: Do you understand what I mean ? And maybe I have some missundend about the action server.

julianoes: Your English is fine, don't worry! The part I don't understand is why you want to implement it differently for the camera server, or why you can't use the same way it is done in the ActionServer? I would like to have the servers consistent/all the same, so that it is easy to implement against.

tbago: Yeah, I think the same way to implement the server is good disign. But for our camera usage, all the camera function and response result is platform dependency, So we can just use the CameraServer as a delegate.
For other camera implement may have the same design, the camera is all running on specail embeded chip and have different sdk.

julianoes: Hmm, I don't really want to make an exception because of some special embedded camera SDK. What's the problem, restriction that you run into? Surely, this is solvable with software? Or can you make a code example to illustrate the problem?

tbago: Hi Julianoes, sorry for long time no replay:
I just draw another sequence diagram to show the basic logic (this logic is already implement in the [MAVSDK](https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/camera_server/camera_server.proto#L18-L21)) 
``` mermaid
sequenceDiagram
    participant c as MAVClient
    participant s as MAVServer
    participant b as Backend(running on embeded chip)
    loop
        b-->>s: subscribe take photo callback
        c-->>s: call take photo
        s-->>b: callback(the pre subscribe method)
        
        Note right of b: process the take photo command
        Note right of b: call embeded chip sdk to really take photo

        alt respond take photo 
            b->>s: Success take photo
        else Failed
            b->>s: Interal embeded system error
        else Busy
            b->>s: Not ready take photo
        end

        s->>c: responed take photo result
    end
  
```

tbago: On more question:
First I am not familiar with the action server, so maybe I have some misunderstand. For example if I have a real drone (running with the action server), and I want to arm and take off the drone, so first I need check the drone state(like battery is ok).
So the action server also need to communicate the drone (running on embeded system) to get state.

julianoes: @JonasVautherin do you have an opinion here? Maybe I'm confused.

julianoes:
@tbago just letting you know: I'm now looking into it and will be working on it to get it merged as soon as possible.

tbago:
Ok, I would like to contribute to the open source community, so what can I do for now?

julianoes:
If it's ok I will force push your branch here with the latest changes I have been doing and we continue from there?

julianoes:
@tbago I've force pushed it. I'll also force push the MAVSDK PR: https://github.com/mavlink/MAVSDK-Proto/pull/320
And then work on merging it.

julianoes (approved):

protos/camera_server/camera_server.proto
julianoes: I think I was confused, and @tbago was right.

JonasVautherin (approved):

PR #321, opened 2023-06-26
Customizing distance sensor to add orientation for multiple sensor support

prokas-nikos:
Added orientation parameter to telemetry.distance_sensor to take advantage of the mavlink orientation parameter and receive measurements for multiple distance sensors

### DIFF
--- a/protos/telemetry/telemetry.proto
+++ b/protos/telemetry/telemetry.proto
@@ -742,6 +742,7 @@ message DistanceSensor {
     float minimum_distance_m = 1 [(mavsdk.options.default_value)="NaN"]; // Minimum distance the sensor can measure, NaN if unknown.
     float maximum_distance_m = 2 [(mavsdk.options.default_value)="NaN"]; // Maximum distance the sensor can measure, NaN if unknown.
     float current_distance_m = 3 [(mavsdk.options.default_value)="NaN"]; // Current distance reading, NaN if unknown.
+    uint32 orientation = 4  [(mavsdk.options.default_value)="0"]; // Sensor Orientation reading, 0 if unknown.
 }
 
 // Scaled Pressure message type.
--- a/protos/telemetry/telemetry.proto
+++ b/protos/telemetry/telemetry.proto
@@ -742,7 +742,7 @@ message DistanceSensor {
     float minimum_distance_m = 1 [(mavsdk.options.default_value)="NaN"]; // Minimum distance the sensor can measure, NaN if unknown.
     float maximum_distance_m = 2 [(mavsdk.options.default_value)="NaN"]; // Maximum distance the sensor can measure, NaN if unknown.
     float current_distance_m = 3 [(mavsdk.options.default_value)="NaN"]; // Current distance reading, NaN if unknown.
-    uint32 orientation = 4  [(mavsdk.options.default_value)="0"]; // Sensor Orientation reading, 0 if unknown.
+    EulerAngle orientation = 4; // Sensor Orientation reading.
 }
 
 // Scaled Pressure message type.


protos/telemetry/telemetry.proto
```diff
@@ -742,6 +742,7 @@ message DistanceSensor {
     float minimum_distance_m = 1 [(mavsdk.options.default_value)="NaN"]; // Minimum distance the sensor can measure, NaN if unknown.
     float maximum_distance_m = 2 [(mavsdk.options.default_value)="NaN"]; // Maximum distance the sensor can measure, NaN if unknown.
     float current_distance_m = 3 [(mavsdk.options.default_value)="NaN"]; // Current distance reading, NaN if unknown.
+    uint32 orientation = 4  [(mavsdk.options.default_value)="0"]; // Sensor Orientation reading, 0 if unknown.
```
JonasVautherin, line 745: What's the unit of that int? Degrees?

julianoes: It's this annoying orientation enum: https://mavlink.io/en/messages/common.html#MAV_SENSOR_ORIENTATION

It's incomplete and usually lacking the angle that you want :see_no_evil:.

JonasVautherin: Oh... Should we translate that to `orientation_deg`?

julianoes: It would have to be 3 angles.

JonasVautherin: Oh, the enum is weirder than I thought :see_no_evil:. Then shouldn't it be an `Euler` or a `Quaternion`?

julianoes: That would be nice, if MAVLink was using euler angles for that rather than an enum, yes.

JonasVautherin: We don't need MAVLink to change: both `Euler` and `Quaternion` can encode whatever is in this weird MAVLink enum, we just have to convert it. But that's an implementation detail, right?

julianoes: Yes, we can do that.

prokas-nikos: So it would be a better idea to make an other pull request with
orientation_pitch = 4
orientation_roll = 5
orientation_yaw = 6
and map the MAV_SENSOR_ORIENTATION in telemetry_impl.cpp with the euler values(pitch, roll, yaw)?

JonasVautherin: Yes, but reusing one of the types of telemetry.proto: `Euler` or `Quaternion`.

prokas-nikos: I have updated the telemetry.proto but the implementation is so ugly now that each enum is mapped to an EulerAngle one by one. It works but the code looks like this.

Telemetry::EulerAngle euler_angle;
    euler_angle.roll_deg = 0;
    euler_angle.pitch_deg = 0;
    euler_angle.yaw_deg = 0;
    
    switch( orientation )
    {
        case MavSensorOrientation::MAV_SENSOR_ROTATION_YAW_45: {
            euler_angle.yaw_deg = 45;
            break;
        }
        case MavSensorOrientation::MAV_SENSOR_ROTATION_YAW_90: {
            euler_angle.yaw_deg = 90;
            break;
        }
        case MavSensorOrientation::MAV_SENSOR_ROTATION_YAW_135: {
            euler_angle.yaw_deg = 135;
            break;
        }
......... in total 44 cases ....

julianoes: I don't think that's ugly, at least it makes the API nicer!

julianoes (approved):

protos/telemetry/telemetry.proto
JonasVautherin: It is ugly because the MAVLink enum is weird :grin:. But I think this is the right thing to do :+1:

JonasVautherin (approved):

PR #322, opened 2023-06-28
feature zoom focus tracking

aminballoon:
From this PR -> https://github.com/mavlink/MAVSDK/pull/2094#issuecomment-1610226707
This repository add the camera function to proto

### DIFF
--- a/protos/camera/camera.proto
+++ b/protos/camera/camera.proto
@@ -116,129 +116,49 @@ service CameraService {
      */
     rpc ResetSettings(ResetSettingsRequest) returns(ResetSettingsResponse) {}
     /*
-     * Zoomin Camera.
-     *
+     * Start zooming in.
      */
-    rpc CameraZoomWide(CameraZoomWideRequest) returns(CameraZoomWideResponse) {}
+    rpc ZoomInStart(ZoomInStartRequest) returns(ZoomInStartResponse) {}
     /*
-     * Zoomout Camera.
-     *
+     * Start zooming out.
      */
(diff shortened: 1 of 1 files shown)


julianoes:
Thanks for the PR. Are you aware of https://github.com/mavlink/MAVSDK-Proto/pull/320? I'm just worried you might run into conflicts given you're both working on camera things.

aminballoon:
Thanks for good news . I have read #320 I'm not worried about this PR. These 2 PR should be able to combine because the functions are not duplicated.

JonasVautherin:
Maybe a question for @julianoes here: how much do those functions overlap with camera params? Is it expected to have an API here for some params, but not all?

aminballoon:
@JonasVautherin I think it's depend on camera model some camera payload doesn't support Mavlink command they may not be able to use this feature either.

julianoes:
@JonasVautherin zoom and focus are not implemented using parameters but standardized commands:
https://mavlink.io/en/messages/common.html#MAV_CMD_SET_CAMERA_ZOOM
https://mavlink.io/en/messages/common.html#MAV_CMD_SET_CAMERA_FOCUS

So I think this is ok.

JonasVautherin:
I know at least two cameras that have zoom and/or focus in the camera definition :innocent:. But yeah I guess that's part of the MAVLink specs: the use should try to detect (or hardcode) the camera and then hardcode the correct behavior :see_no_evil:.

julianoes:
Yer zoom was added later, as was focus.

julianoes (approved):
I've edited this a bit. I added the units, and removed the redundant camera prefix.

JonasVautherin (approved):

PR #323, opened 2023-08-15
mission_raw: add planned home position

JonasVautherin:
Goes with https://github.com/mavlink/MAVSDK/pull/2115

The thing is that now it forces everybody to set that value (we can't have it optional with the auto-generation) :thinking:. I wonder if that's reasonable or not :thinking:.

Do users even use the `MissionImportData` directly, or is it just used in the implementation somehow?

@RomanBapst: what do you think?

### DIFF
--- a/protos/mission_raw/mission_raw.proto
+++ b/protos/mission_raw/mission_raw.proto
@@ -170,6 +170,8 @@ message MissionImportData {
     repeated MissionItem mission_items = 1; // Mission items
     repeated MissionItem geofence_items = 2; // Geofence items
     repeated MissionItem rally_items = 3; // Rally items
+    bool has_planned_home = 4; // Whether it contains a planned home position element or not
+    MissionItem planned_home_position = 5; // Planned home position
 }
 
 // Result type.


JonasVautherin:
Or maybe it should be done in a separate function, such that it is optional. Like we do in the mission plugin: https://github.com/mavlink/MAVSDK-Proto/blob/main/protos/mission/mission.proto#L104

JonasVautherin:
> The thing is that now it forces everybody to set that value

Ok now it has a `has_planned_home_position` boolean to say whether that field is set or not :+1:.

julianoes (commented):
But this won't have any effect for PX4, which would be confusing.

Plus, from what I hear this home position is mostly not used for ArduPilot.

Or maybe I'm misunderstanding this. Are we talking about the ArduPilot feature where home needs to be in 0, or are we talking about a general functionality that allows a mission to be moved to where the vehicle is?

JonasVautherin:
> But this won't have any effect for PX4, which would be confusing.

I think it is just part of the parsing of a QGC `.plan`, in this case made for PX4 :thinking:. My understanding is that we already do parse it, it's just that we ignore this element (and therefore lose it in translation). In this case Roman needs this element for the PX4 integration tests, I believe.

RomanBapst:
@julianoes For the PX4 integration tests I need to be able to move the mission to the location of the vehicle.
We already do this [here](https://github.com/PX4/PX4-Autopilot/blob/main/test/mavsdk_tests/test_vtol_mission.cpp#L42)
However, what we do is to move the mission such that the first waypoint coincides with the vehicle position. For a VTOL mission this is not correct as the first waypoint is a VTOL takeoff which actually specifies a transition direction and should not be places over the vehicle.
That's why I needed to get hold of the planned home position which I saw MAVSDK did not expose yet and thus the PR.

RomanBapst:
> So what's the behavior if someone sets that field and then the home position is different from the planned one? Is everything shifted?

@julianoes This is not really the problem of MAVSDK, is it? All that MAVSDK does it telling the user that the plan file which has been loaded contains a planned home position and the values are such and such. The integrator can use that (I am using is for integration testing) but MAVSDK should not care. Does that make sense?

julianoes:
Oh, now I understand! This has nothing to do with upload. Sorry, I missed that. Hm, ok. Would be nice to have an example that shows how to use the value, if set. Or at least make it very clear from the description of the field. And make sure to fix the linting/CI.

julianoes:
This won't go into v1.4 though but I'd say v2/main.

RomanBapst:
@julianoes I will try to add an example. Maybe loading a VTOL mission from a plan file and making sure that the distance between planned home and the VTOL takeoff item is more than X?

julianoes:
Sure, whatever you would use it for.

PR #327, opened 2024-01-11
remote_id: open drone id plugin interface

persuader72:
Proto file for the [remote_id](https://mavlink.io/en/services/opendroneid.html) protocol.

This proto file is need by the new [remote_id](https://github.com/persuader72/MAVSDK/tree/remote_id) MAVSDK plugin.

The goal of the plugin to use the companion computer to work with [open drone id](https://www.opendroneid.org/) compatible hardware like [this](https://github.com/ArduPilot/ArduRemoteID) even is the autopilot is not compatible with such protocol.


protos/remote_id/remote_id.proto
```diff
@@ -0,0 +1,125 @@
...
+
+message SetSelfIdRequest {
+    SelfId self_id = 1; // Desidered self_id
+}
+message SetSelfIdResponse {
+    RemoteIdResult remote_id_result = 1;
+}
+
+message BasicId {
+    uint32 id_type= 1;  // Indicates the format for the uas_id field of this message.
```
JonasVautherin, line 69:
```suggestion
    uint32 id_type = 1;  // Indicates the format for the uas_id field of this message.
```

protos/remote_id/remote_id.proto
```diff
@@ -0,0 +1,125 @@
...
+    uint32 ua_type = 2; // Indicates the type of UA (Unmanned Aircraft).
+    string uas_id = 3;  // UAS ID following the format specified by id_type.
+}
+
+message Location {
+    uint32 status= 1;             // Indicates whether the unmanned aircraft is on the ground or in the air.
+    uint32 direction= 2;          // Direction over ground measured clockwise from true North: 0 - 35999 cdeg.
+    uint32 speed_horizontal= 3;   // Ground speed. Positive only. cm/s
```
JonasVautherin, line 77: Our convention is that we add the unit as a suffix, e.g:

```suggestion
    uint32 speed_horizontal_cm_s= 3;   // Ground speed. Positive only. cm/s
```

Then my second question would be: can't it be `m/s`? Feels more natural to me for an aircraft :thinking:

protos/remote_id/remote_id.proto
```diff
@@ -0,0 +1,125 @@
...
+package mavsdk.rpc.remote_id;
+
+import "mavsdk_options.proto";
+
+option java_package = "io.mavsdk.remote_id";
+option java_outer_classname = "RemoteIdProto";
+
+service RemoteIdService {
+    /*
+     * Update the BasicId strucutre sent with the basic_id packet
```
JonasVautherin, line 12:
```suggestion
     * Update the BasicId structure sent with the basic_id packet
```

JonasVautherin (changes requested):

protos/remote_id/remote_id.proto
```diff
@@ -0,0 +1,125 @@
...
+    uint32 id_type = 1; // Indicates the format for the uas_id field of this message.
+    uint32 ua_type = 2; // Indicates the type of UA (Unmanned Aircraft).
+    string uas_id = 3; // UAS ID following the format specified by id_type.
+}
+
+message Location {
+    uint32 status = 1; // Indicates whether the unmanned aircraft is on the ground or in the air.
+    uint32 direction_deg = 2; // Direction over ground measured clockwise from true North: 0 - 359 deg.
+    uint32 speed_horizontal_cm_s = 3; // Ground speed. Positive only. cm/s
+    int32 speed_vertical_cm_s = 4; // The vertical speed. Up is positive. cm/s
```
JonasVautherin, line 78: Thanks for the changes :blush:.

Is there a reason to have this in `cm/s` instead of `m/s`?

protos/remote_id/remote_id.proto
```diff
@@ -0,0 +1,125 @@
...
+    float latitude_deg = 5; // Current latitude of the unmanned aircraft
+    float longitude_deg = 6; // Current longitude of the unmanned aircraft
```
julianoes, line 80: Lat/lon need to have `double` precision.

protos/remote_id/remote_id.proto
```diff
@@ -0,0 +1,125 @@
...
+    float operator_latitude_deg = 3; // Latitude of the operator.
+    float operator_longitude_deg = 4; // Longitude of the operator.
+    uint32 area_count = 5; // Number of aircraft in the area,
+    uint32 area_radius_m = 6; // Radius of the cylindrical area of the group or formation.
+    float area_ceiling_m = 7; // Area Operations Ceiling relative to WGS84.
+    float area_floor_m = 8; // Area Operations Floor relative to WGS84.
+    uint32 category_eu = 9; // When classification_type is MAV_ODID_CLASSIFICATION_TYPE_EU, specifies the category of the UA.
+    uint32 class_eu = 10; // When classification_type is MAV_ODID_CLASSIFICATION_TYPE_EU, specifies the class of the UA.
+    float operator_altitude_geo_m = 11; // Geodetic altitude of the operator relative to WGS84. If unknown: -1000 m.
+    uint32 timestamp_s = 12; // s32 bit Unix Timestamp in seconds since 00:00:00 01/01/2019.
```
julianoes, line 101: Wow, they came up with some weird types in the opendrone ID MAVLink messages. A UNIX timestamp but starting in 2019. It'll overflow in year 2155. Well, won't be my problem :joy:

julianoes: I wonder if we want to expose this or if MAVSDK should read the time from the system and avoid having the user to deal with it.

protos/remote_id/remote_id.proto
```diff
@@ -0,0 +1,125 @@
...
+}
+
+message OperatorId {
+    uint32 operator_id_type = 1; // Indicates the type of the description field.
```
julianoes, line 105: Would be good if all the types are proper enums.

protos/remote_id/remote_id.proto
```diff
@@ -0,0 +1,125 @@
...
+    uint32 speed_horizontal_cm_s = 3; // Ground speed. Positive only. cm/s
+    int32 speed_vertical_cm_s = 4; // The vertical speed. Up is positive. cm/s
+    float latitude_deg = 5; // Current latitude of the unmanned aircraft
+    float longitude_deg = 6; // Current longitude of the unmanned aircraft
+    float altitude_barometric_m = 7; // The altitude calculated from the barometric pressure.
+    float altitude_geodetic_m = 8; // The geodetic altitude as defined by WGS84.
+    uint32 height_reference = 9; // Indicates the reference point for the height field.
+    float height_m = 10; // The current height of the unmanned aircraft. As indicated by height_reference.
+    float timestamp_s = 11; // Seconds after the full hour with reference to UTC time.
+    uint32 timestamp_accuracy = 12; // The accuracy of the timestamps.
```
julianoes, line 86: Gosh, this is again so ugly in the MAVLink message. It's an enum in steps of 0.1s to signal accuracy. I wonder who thought this would be a good idea...

In any case, my suggestion would be to make the API human usable by just doing `timestamp_accuracy_s`, so in seconds, and then do the ugly conversion inside MAVSDK.

julianoes (commented):
Thanks for your PR @persuader72. I'm appalled by the OpenDrone ID MAVLink messages, now that I look at them in detail but a well, too late now, I suppose.

However, we have to figure out which things we want to expose raw and what we can hide and make easier.

protos/remote_id/remote_id.proto
JonasVautherin: > if MAVSDK should read the time from the system

I agree it would be nice, but I'm scared of it. MAVSDK runs on different platforms, and reading the time may not be as straightforward as we may think? Just a thought, I don't have a strong opinion there.

julianoes: Yes, that's potentially an issue. Also local time or UTC? :thinking:

protos/remote_id/remote_id.proto
persuader72: laziness I think :flushed:. 
I used the same units presents in  mavlink packets. 
I agree that is better to use SI units and let he plugin handle the conversions.

protos/remote_id/remote_id.proto
persuader72: The time is UTC but:
I'm experimenting with the plugin and I tried to fill the timestamp for the user but the problem I see is that a  reliable clock source in a embedded system, maybe, is a requirement too strong for the library.
Instead the precision of clock is important because the remoteid standard said that packets must be rejected if they are more old than one second. 
So in the end I think that is better let user get the responsibility to deal with timestamps.

persuader72: Maybe, We can let user provide a valid UTC timestamp and let plugin make the weird conversions?

julianoes: I'd say so

protos/remote_id/remote_id.proto
persuader72: i think the the suggested changes are completed.

protos/remote_id/remote_id.proto
```diff
@@ -0,0 +1,275 @@
...
+
+    enum HeightRef {
+        HEIGHT_REF_OVER_TAKEOFF = 0; // The height field is relative to the take-off location.
+        HEIGHT_REF_OVER_GROUND = 1; // The height field is relative to ground.
+    }
+
+    Status status = 1; // Indicates whether the unmanned aircraft is on the ground or in the air.
+    uint32 direction_deg = 2; // Direction over ground measured clockwise from true North: 0 - 359 deg.
+    float speed_horizontal_m_s = 3; // Ground speed. Positive only. m
+    float speed_vertical_m_s = 4; // The vertical speed. Up is positive. m
```
JonasVautherin, line 129:
```suggestion
    float speed_horizontal_m_s = 3; // Ground speed in meters per second. Positive only.
    float speed_vertical_m_s = 4; // The vertical speed in meters per second. Up is positive.
```

protos/remote_id/remote_id.proto
```diff
@@ -0,0 +1,275 @@
...
+        CLASS_EU_CLASS_4 = 5; // The class for the UA, according to the EU specification, is Class 4.
+        CLASS_EU_CLASS_5 = 6; // The class for the UA, according to the EU specification, is Class 5.
+        CLASS_EU_CLASS_6 = 7; // The class for the UA, according to the EU specification, is Class 6.
+}
+
+    OperatorLocationType operator_location_type = 1; // Specifies the operator location type.
+    ClassificationType classification_type = 2; // Specifies the classification type of the UA.
+    double operator_latitude_deg = 3; // Latitude of the operator.
+    double operator_longitude_deg = 4; // Longitude of the operator.
+    uint32 area_count = 5; // Number of aircraft in the area,
```
JonasVautherin, line 234:
```suggestion
    uint32 area_count = 5; // Number of aircraft in the area.
```

protos/remote_id/remote_id.proto
```diff
@@ -0,0 +1,275 @@
...
+
+    enum UasType {
+        UAS_TYPE_NONE = 0; // No UA (Unmanned Aircraft) type defined.
+        UAS_TYPE_AEROPLANE = 1; // Aeroplane/Airplane. Fixed wing.
+        UAS_TYPE_HELICOPTER_OR_MULTIROTOR = 2; // Helicopter or multirotor.
+        UAS_TYPE_GYROPLANE = 3; // Gyroplane.
+        UAS_TYPE_HYBRID_LIFT = 4; // VTOL (Vertical Take-Off and Landing). Fixed wing aircraft that can take off vertically.
+        UAS_TYPE_ORNITHOPTER = 5; // Ornithopter.
+        UAS_TYPE_GLIDER = 6; // Glider.
+        UAS_TYPE_KIT = 7; // Kite.
```
JonasVautherin, line 96: Shouldn't it be `UAS_TYPE_KITE`?

JonasVautherin (commented):
A few nitpicks :innocent:.

persuader72:
With the "ArmStatus" stream i think that the feature set is completed

protos/remote_id/remote_id.proto
```diff
@@ -0,0 +1,294 @@
...
+message ArmStatusResponse {
+    ArmStatus arm_status = 1;
+}
+
+message BasicId {
+    enum IdType {
+        ID_TYPE_NONE = 0; // No type defined.
+        ID_TYPE_SERIAL_NUMBER = 1; // Manufacturer Serial Number (ANSI/CTA-2063 format).
+	    ID_TYPE_CAA_REGISTRATION_ID = 2; // CAA (Civil Aviation Authority) registered ID. Format: [ICAO Country Code].[CAA Assigned ID].
+	    ID_TYPE_UTM_ASSIGNED_UUID = 3; // UTM (Unmanned Traffic Management) assigned UUID (RFC4122).
```
JonasVautherin, line 93:
```suggestion
        ID_TYPE_CAA_REGISTRATION_ID = 2; // CAA (Civil Aviation Authority) registered ID. Format: [ICAO Country Code].[CAA Assigned ID].
        ID_TYPE_UTM_ASSIGNED_UUID = 3; // UTM (Unmanned Traffic Management) assigned UUID (RFC4122).
```

protos/remote_id/remote_id.proto
```diff
@@ -0,0 +1,294 @@
...
+
+message SetBasicIdRequest {
+    BasicId basic_id = 1; // Desired basic_id
+}
+message SetBasicIdResponse {
+    RemoteIdResult remote_id_result = 1;
+}
+
+message SetLocationRequest {
+    Location location = 1; // Desidered location
```
JonasVautherin, line 49: There are a few typos :see_no_evil: 

```suggestion
    Location location = 1; // Desired location
```

protos/remote_id/remote_id.proto
```diff
@@ -0,0 +1,294 @@
...
+
+    enum VerAcc {
+        VER_ACC_UNKNOWN = 0; // The vertical accuracy is unknown.
+        VER_ACC_METER_150 = 1; // The vertical accuracy is smaller than 150 meter.
+        VER_ACC_METER_45 = 2; // The vertical accuracy is smaller than 45 meter.
+        VER_ACC_METER_25 = 3; // The vertical accuracy is smaller than 25 meter.
+        VER_ACC_METER_10 = 4; // The vertical accuracy is smaller than 10 meter.
+        VER_ACC_METER_3 = 5; // The vertical accuracy is smaller than 3 meter.
+        VER_ACC_METER_1 = 6; // The vertical accuracy is smaller than 1 meter.
+    }
```
julianoes, line 173: Still puzzles my why they wouldn't just use a `uint8` to signal meter accuracy... but anyway, I think I can live with that.

julianoes (approved):
I think this looks persuading :smile:

protos/remote_id/remote_id.proto
persuader72: I think the reason is historic. In some radio broadcasting ID protocols (like AIS or ADSB) the accuracy is an enumerative type or is rounded to less than 8 bits to save space (then save bandwidth)

persuader72:
@julianoes, It is better to rewrite the history of this PR in one single commit?

protos/remote_id/remote_id.proto
julianoes: Right, fair enough.

julianoes:
@persuader72 Feel free to rewrite the history and force push. If you don't, then I'll just squash-merge it.

julianoes:
Sorry, this disappeared off my radar. I will review again ASAP!

josephm28:
@julianoes @persuader72 Is it possible to merge this?

julianoes:
Sorry yes, I need to work through this. I have it planned as soon as v3 is done (see https://github.com/mavlink/MAVSDK/issues/2316).

aheyne:
@julianoes Now that v3 is done any chance you have cycles for this again?

julianoes:
Might as well, thanks for the reminder! I'll review tomorrow.

julianoes (approved):
I think I can live with this.

julianoes:
@persuader72 are you still interested in finishing this?

aheyne:
@julianoes It's not clear to me what's needed to get this over the line but I would be interested in finishing the work if you can provide some guidance.

persuader72:
Sorry for the late reply , yes In the next days I have some time to spend in this activity, 

@aheyne if you want we can split the work to finish the module.

julianoes:
@persuader72 can you make a pull request for the C++ MAVSDK part?

julianoes:
Continuing in #387.

PR #328, opened 2024-01-12
Full gimbal angle control

Daybreakerflint:
This feature is a request as I have to use the gimbal I have in a certain way and the roll and pitch axis are switched.
As the pitch of the payload of the gimbal is vitally important, this feature is needed.

### DIFF
--- a/protos/gimbal/gimbal.proto
+++ b/protos/gimbal/gimbal.proto
@@ -141,7 +141,7 @@ message ControlResponse {
 // Gimbal mode type.
 enum GimbalMode {
     GIMBAL_MODE_YAW_FOLLOW = 0; // Yaw follow will point the gimbal to the vehicle heading
-    GIMBAL_MODE_YAW_LOCK = 1; // Yaw lock will fix the gimbal poiting to an absolute direction
+    GIMBAL_MODE_YAW_LOCK = 1; // Yaw lock will fix the gimbal pointing to an absolute direction
 }
 
 // Control mode
--- a/protos/gimbal/gimbal.proto
+++ b/protos/gimbal/gimbal.proto
@@ -9,9 +9,9 @@ option java_outer_classname = "GimbalProto";
 service GimbalService {
     /*
      *
-     * Set gimbal pitch, yaw and roll angles.
+     * Set gimbal roll, pitch and yaw angles.
      *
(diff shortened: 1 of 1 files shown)


JonasVautherin:
> and the roll and pitch axis are switched

Can you elaborate a bit? I am not sure I understand what this means... Are you saying that your Gimbal is reading the MAVLink message incorrectly, and therefore you want to invert the values you send from MAVSDK instead of fixing the Gimbal side?

Daybreakerflint:
Hey Jonas!
I do not just want to invert the control, as my setup needs it because how i have it the gimbal is balanced and I have the least amount of weight. 

My payload on the gimbal is a quad rigged horn antenna. An antenna with two polarization. 
Because the drone payload is sensitive to such things as alignment. Meaning also how the polarization is to the object.
IMG_20240109_153559

The gimbal is reading everything ok. I just want full control of where it points to.

JonasVautherin:
I see, so you need the roll axis. It was initially not exposed because cameras don't really need to set the roll axis :innocent:.

@julianoes: do you have an opinion on that?

protos/gimbal/gimbal.proto
```diff
@@ -71,6 +80,15 @@ service GimbalService {
      rpc SubscribeControl(SubscribeControlRequest) returns(stream ControlResponse) {}
 }
 
+message SetAnglesRequest {
+    float pitch_deg = 1; // Pitch angle in degrees (negative points down)
+    float yaw_deg = 2; // Yaw angle in degrees (positive is clock-wise, range: -180 to 180 or 0 to 360)
+    float roll_deg = 3; // Roll angle in degrees
```
julianoes, line 86: Can we please change the order of the fields to match the usual MAVLink order: roll, pitch, yaw.

julianoes (changes requested):

protos/gimbal/gimbal.proto
Daybreakerflint: Sure. Shall I do it, or do you do it?  As I said, this is my first time. Happy to help where I can.

julianoes: I suggest you change it. Thanks!

JonasVautherin (approved):

julianoes (approved):

PR #330, opened 2024-02-22
support multi video stream info and add camera cap flags

tbago:
This PR contain the following features:

- The camera proto video stream info contain more detail about the stream info, so the camera server proto need use the same video stream info define;
- the camera can have multi lens (e.g. one Visible light one Infrared). So need support multi video stream info;
- the camera cap flags is important for clint to check the camera capability;
- unify the camera information int camera and camera server;

### DIFF
--- a/protos/camera/camera.proto
+++ b/protos/camera/camera.proto
@@ -422,11 +422,50 @@ message SettingOptions {
 
 // Type to represent a camera information.
 message Information {
-    string vendor_name = 1; // Name of the camera vendor
-    string model_name = 2; // Name of the camera model
-    float focal_length_mm = 3; // Focal length
-    float horizontal_sensor_size_mm = 4; // Horizontal sensor size
-    float vertical_sensor_size_mm = 5; // Vertical sensor size
-    uint32 horizontal_resolution_px = 6; // Horizontal image resolution in pixels
-    uint32 vertical_resolution_px = 7; // Vertical image resolution in pixels
+    enum CameraCapFlags {
+        CAMERA_CAP_FLAGS_CAPTURE_VIDEO = 0;  // Camera is able to record video
+        CAMERA_CAP_FLAGS_CAPTURE_IMAGE = 1;  // Camera is able to capture images
+        CAMERA_CAP_FLAGS_HAS_MODES =
+            2;  // Camera has separate Video and Image/Photo modes
+                // (MAV_CMD_SET_CAMERA_MODE)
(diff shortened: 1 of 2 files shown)


tbago:
The pull request to MAVSDK will come soon

protos/camera/camera.proto
```diff
@@ -421,11 +422,50 @@ message SettingOptions {
...
+                // video streaming info)
+        CAMERA_CAP_FLAGS_HAS_TRACKING_POINT =
+            9;  // Camera supports tracking of a point on the camera view.
+        CAMERA_CAP_FLAGS_HAS_TRACKING_RECTANGLE =
+            10;  // Camera supports tracking of a selection rectangle on the
+                 // camera view.
+        CAMERA_CAP_FLAGS_HAS_TRACKING_GEO_STATUS =
+            11;  // Camera supports tracking geo status
+                 // (CAMERA_TRACKING_GEO_STATUS).
+    };
```
julianoes, line 453: The capture flags should only be internally used. They should not be visible to the user. And if we need to expose specific ones, then they have to just be boolean flags.

protos/camera/camera.proto
```diff
@@ -421,11 +422,50 @@ message SettingOptions {
...
+    float horizontal_sensor_size_mm = 5;  // Horizontal sensor size
+    float vertical_sensor_size_mm = 6;    // Vertical sensor size
+    uint32 horizontal_resolution_px =
+        7;                              // Horizontal image resolution in pixels
+    uint32 vertical_resolution_px = 8;  // Vertical image resolution in pixels
+    uint32 lens_id = 9;                 // Lens ID
+    uint32 definition_file_version =
+        10;  // Camera definition file version (iteration)
+    string definition_file_uri =
+        11;  // Camera definition URI (http or mavlink ftp)
```
julianoes, line 468: This is also only to be used internally.

protos/camera/camera.proto
```diff
@@ -190,7 +190,7 @@ message ModeResponse {
 
 message SubscribeVideoStreamInfoRequest {}
 message VideoStreamInfoResponse {
-    VideoStreamInfo video_stream_info = 1; // Video stream info
+    repeated VideoStreamInfo video_stream_infos = 1; // Video stream infos
```
julianoes, line 193: How does this work when it's repeated? Do you get multiple video stream infos, one per video stream?

julianoes (changes requested):

PR #332, opened 2024-03-21
Create winch_reel.proto

jperez-droneup:
Added a winch_reel plugin based on the existing mavlink-compliant winch plugin. This plugin contains only the commands supported by Watts Reel winch, and includes telemetry specific to the Reel winch. Specifically, the bitmask `status` is reinterpreted as an enum value.

The list of commands and telemetry values includes are those I have personally discovered via company internal communication. If there is a public documentation of these commands, I would be happy to link it in here.

### DIFF
--- a/protos/winch_reel/winch_reel.proto
+++ b/protos/winch_reel/winch_reel.proto
@@ -1,0 +1,177 @@
+syntax = "proto3";
+
+package mavsdk.rpc.winch_reel;
+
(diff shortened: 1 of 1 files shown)


julianoes:
As discussed via email, MAVSDK can't carry workarounds for the Watts reel-winch and instead the Watts product should be updated to adhere to the MAVLink spec.

PR #333, opened 2024-03-21
camera_server: implement zoom

julianoes:
This implements continuous and range zooming.


julianoes (commented):
Self reviewed with fresh eyes.

PR #334, opened 2024-04-01
Added arm authorization server functionality

ddatsko:
This PR adds the functionality of arm authorization server, as it is described by [MavLink arm authorization protocol](https://mavlink.io/en/services/arm_authorization.html). The server should be able to respond to `MAV_CMD_ARM_AUTHORIZATION_REQUEST` with a user specified callback and allow or reject arming, requested by a vehicle. I also created a [PR into MAVSDK](https://github.com/mavlink/MAVSDK/pull/2267) with a C++ implementation and example

### DIFF
--- a/protos/arm_authorizer_server/arm_authorizer_server.proto
+++ b/protos/arm_authorizer_server/arm_authorizer_server.proto
@@ -26,12 +26,12 @@ message SubscribeArmAuthorizationResponse {
 
 // Messages for RespondArmAuthorization
 message AcceptArmAuthorizationRequest {
-    int32 valid_time = 1; // Time in seconds for which this authorization is valid
+    int32 valid_time_s = 1; // Time in seconds for which this authorization is valid
 }
 
 // Result type
 message AcceptArmAuthorizationResponse {
-    CommandAnswer command_answer = 1; // Result enum value
+    ArmAutorizerServerResult arm_authorizer_server_result = 1; // Result enum value
 }
 
 // Messages for RespondArmAuthorization
@@ -43,12 +43,17 @@ message RejectArmAuthorizationRequest {
 
 // Result type
 message RejectArmAuthorizationResponse {
-    CommandAnswer command_answer = 1; // Result enum value
+    ArmAutorizerServerResult arm_authorizer_server_result = 1; // Result enum value
 }
 
-enum CommandAnswer {
-    COMMAND_ANSWER_ACCEPTED = 0; // Command accepted
-    COMMAND_ANSWER_FAILED = 1; // Command failed
+message ArmAutorizerServerResult {
+    enum Result {
+        RESULT_SUCCESS = 0; // Command accepted
+        RESULT_FAILED = 1; // Command failed
+    }
+
+    Result result = 1; // Result enum value
+    string result_str = 2; // Human-readable English string describing the result
 }
 
 enum RejectionReason {
--- a/protos/arm_authorizer_server/arm_authorizer_server.proto
+++ b/protos/arm_authorizer_server/arm_authorizer_server.proto
@@ -57,10 +57,10 @@ message ArmAutorizerServerResult {
 }
 
 enum RejectionReason {
-    REASON_GENERIC = 0; // Not a specific reason
-    REASON_NONE = 1; // Authorizer will send the error as string to GCS
-    REASON_INVALID_WAYPOINT = 2; // At least one waypoint have a invalid value
-    REASON_TIMEOUT = 3; // Timeout in the authorizer process(in case it depends on network)
-    REASON_AIRSPACE_IN_USE = 4; // Airspace of the mission in use by another vehicle, second result parameter can have the waypoint id that caused it to be denied.
-    REASON_BAD_WEATHER = 5; // Weather is not good to fly
+    REJECTION_REASON_GENERIC = 0; // Not a specific reason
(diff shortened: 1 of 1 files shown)


protos/arm_authorizer_server/arm_authorizer_server.proto
```diff
@@ -0,0 +1,61 @@
...
+
+// Messages for SubscribeArmAuthorization
+message SubscribeArmAuthorizationRequest {}
+message SubscribeArmAuthorizationResponse {
+    uint32 system_id = 1; // vehicle system id
+}
+
+// Messages for RespondArmAuthorization
+message AcceptArmAuthorizationRequest {
+    int32 valid_time = 1; // Time in seconds for which this authorization is valid
```
julianoes, line 29: We should specify unit of that time.

protos/arm_authorizer_server/arm_authorizer_server.proto
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// Result type
+message RejectArmAuthorizationResponse {
+    CommandAnswer command_answer = 1; // Result enum value
+}
+
+enum CommandAnswer {
+    COMMAND_ANSWER_ACCEPTED = 0; // Command accepted
+    COMMAND_ANSWER_FAILED = 1; // Command failed
```
julianoes, line 51: I think we're meant to always have the return type `Result`. I'm surprised if it works without that.

protos/arm_authorizer_server/arm_authorizer_server.proto
```diff
@@ -0,0 +1,61 @@
...
+// Result type
+message AcceptArmAuthorizationResponse {
+    CommandAnswer command_answer = 1; // Result enum value
+}
+
+// Messages for RespondArmAuthorization
+message RejectArmAuthorizationRequest {
+    bool temporarily = 1; // True if the answer should be TEMPORARILY_REJECTED, false for DENIED
+    RejectionReason reason = 2; // Reason for the arm to be rejected
+    int32 extra_info = 3; // Extra information specific to the rejection reason (see https://mavlink.io/en/services/arm_authorization.html)
```
julianoes, line 41: I don't find this extra info? Can you help me understand how that works?

ddatsko: I'm not sure if `extra_info` is the best name for this field, actually. In the [arm authorization protocol description in MavLink](https://mavlink.io/en/services/arm_authorization.html), it is stated that `result_param2` field of the ACK message should be set this way:
>  result_param2=if result is ACCEPTED the it should be set with the time in seconds that this authorization is valid otherwise an aditional information about why it was denied should be set. example: for result_param1=MAV_ARM_AUTH_DENIED_REASON_INVALID_WAYPOINT or MAV_ARM_AUTH_DENIED_REASON_AIRSPACE_IN_USE it may have the index of the waypoint that caused it to be denied.

This is pretty ambiguous and not well defined, so I thought we could leave it to the user. So `result_param2` should be filled with the time interval in `AcceptArmAuthorization` function, but with this `extra_info` othersiwe. I can also see that in `px4` firmware, it is checked only if the reason is `MAV_ARM_AUTH_DENIED_REASON_INVALID_WAYPOINT`, exactly as descried in the example above

protos/arm_authorizer_server/arm_authorizer_server.proto
ddatsko: Hmm, It was working this way, but there were some problems with slightly different namings. I have now replaced the return type of `respond...` functions to be the same as in other plugins to keep it consistent

protos/arm_authorizer_server/arm_authorizer_server.proto
julianoes: Ah, got it.

protos/arm_authorizer_server/arm_authorizer_server.proto
```diff
@@ -0,0 +1,66 @@
...
+    string result_str = 2; // Human-readable English string describing the result
+}
+
+enum RejectionReason {
+    REASON_GENERIC = 0; // Not a specific reason
+    REASON_NONE = 1; // Authorizer will send the error as string to GCS
+    REASON_INVALID_WAYPOINT = 2; // At least one waypoint have a invalid value
+    REASON_TIMEOUT = 3; // Timeout in the authorizer process(in case it depends on network)
+    REASON_AIRSPACE_IN_USE = 4; // Airspace of the mission in use by another vehicle, second result parameter can have the waypoint id that caused it to be denied.
+    REASON_BAD_WEATHER = 5; // Weather is not good to fly
```
julianoes, line 65: I think we need to namespace it fully:

```suggestion
    REJECTION_REASON_GENERIC = 0; // Not a specific reason
    REJECTION_REASON_NONE = 1; // Authorizer will send the error as string to GCS
    REJECTION_REASON_INVALID_WAYPOINT = 2; // At least one waypoint have a invalid value
    REJECTION_REASON_TIMEOUT = 3; // Timeout in the authorizer process(in case it depends on network)
    REJECTION_REASON_AIRSPACE_IN_USE = 4; // Airspace of the mission in use by another vehicle, second result parameter can have the waypoint id that caused it to be denied.
    REJECTION_REASON_BAD_WEATHER = 5; // Weather is not good to fly
```

julianoes (approved):

PR #338, opened 2024-04-11
add component_metadata and component_metadata_server

bkueng:
This is for https://github.com/mavlink/MAVSDK/pull/2272


protos/component_metadata/component_metadata.proto
```diff
@@ -0,0 +1,75 @@
...
+option java_package = "io.mavsdk.component_metadata";
+option java_outer_classname = "ComponentMetadataProto";
+
+// Access component metadata json definitions, such as parameters.
+service ComponentMetadataService {
+
+    /*
+     * Request metadata from a specific component
+     */
+    rpc RequestComponent(RequestComponentRequest) returns(EmptyResponse) { option (mavsdk.options.async_type) = SYNC; }
```
julianoes, line 16: Pretty sure they need to be named a very specific way:

```suggestion
    rpc RequestComponent(RequestComponentRequest) returns(RequestComponentResponse) { option (mavsdk.options.async_type) = SYNC; }
```

protos/component_metadata/component_metadata.proto
```diff
@@ -0,0 +1,75 @@
...
+
+// Metadata response
+message MetadataData {
+    bool success = 1; // true if the metadata is available
+    string json_metadata = 2; // The JSON metadata
+}
+
+message EmptyRequest {}
+
+message EmptyResponse {}
```
julianoes, line 55: Every plugin usually has a Result type that is part of the auto-generation design.

I think you need to return a result for the requests. Best case the result actually tells us if we successfully received the metadata that we requested.

If we don't know which types are populated we could return a response like this:

```
message MetadataResponse {
    Result result = 1;
    bool has_parameter_metadata = 2;
    string parameter_metadata_json = 3;
    ...
}

File: protos/component_metadata/component_metadata.proto [hunk 3]
```diff
@@ -0,0 +1,75 @@
...
+
+    /*
+     * Request metadata from the autopilot component
+     */
+    rpc RequestAutopilotComponent(EmptyRequest) returns(EmptyResponse) { option (mavsdk.options.async_type) = SYNC; }
+
+    /*
+     * Register a callback that gets called when metadata is available
+     */
+    rpc SubscribeMetadataAvailable(MetadataAvailableRequest) returns(stream MetadataUpdateResponse) { option (mavsdk.options.async_type) = ASYNC; }
```
julianoes, line 26: Subscriptions are meant to be for things that get updated over time, rather than one time request/response. What is this one meant to be?

protos/component_metadata_server/component_metadata_server.proto
```diff
@@ -0,0 +1,32 @@
...
+
+import "mavsdk_options.proto";
+
+option java_package = "io.mavsdk.component_metadata_server";
+option java_outer_classname = "ComponentMetadataServerProto";
+
+// Provide component metadata json definitions, such as parameters.
+service ComponentMetadataServerService {
+    /*
+     * Provide metadata (can only be called once)
```
julianoes, line 13: Why only once? Does that mean it can't change?

bkueng: There's a protocol extension proposal for dynamic changes, but for now it's not possible. In the majority of cases it's also not needed.

protos/component_metadata/component_metadata.proto
bkueng: Yes I meant it to be a subscription:
- each json type is returned immediately when available
- the same type could change over time (e.g. when rebooting the FC after an update, or dynamically, which is not implemented, but this API makes it possible)

julianoes: Right, that's true. In this case it should actually just be the subscription, without the request. So you subscribe and then you get updates as they come in.

protos/component_metadata/component_metadata.proto
bkueng: From what I could tell it did not make a difference, only the Response suffix is imporant. I changed it now though.

protos/component_metadata/component_metadata.proto
bkueng: I added the result and clarified what the request is for. Initially I had the requested components in the constructor, but I don't think that's possible through the proto files. Thus it's a different method now.

There's no sync API that blocks and waits for metadata. I'm not sure if that's needed?

protos/component_metadata/component_metadata.proto
julianoes: If I remember right, some things don't matter for C++ but might for some language bindings.

protos/component_metadata/component_metadata.proto
julianoes: That's up to you if you want to expose a sync API.

protos/component_metadata/component_metadata.proto
```diff
@@ -0,0 +1,90 @@
...
+message MetadataData {
+    string json_metadata = 1; // The JSON metadata
+}
+
+// Result type.
+message MetadataResult {
+    // Possible results returned for GetMetadata
+    enum Result {
+        RESULT_SUCCESS = 0; // Success
+        RESULT_NOT_AVAILABLE = 1; // Metadata is not available
```
julianoes, line 61: I would think you also need `CONNECTION_ERROR`, `UNSUPPORTED`, `DENIED`, `FAILED` in case the `request_message` call fails with a nack.

protos/component_metadata/component_metadata.proto
```diff
@@ -0,0 +1,90 @@
...
+message MetadataUpdate {
+    uint32 compid = 1; // The component ID
+    MetadataType type = 2; // The metadata type
+    string json_metadata = 3; // The JSON metadata
+}
+
+enum MetadataType {
+    PARAMETER = 0; // Parameter metadata
+    EVENTS = 1; // Event definitions
+    ACTUATORS = 2; // Actuator definitions
```
julianoes, line 89: I think this needs to be:

```suggestion
    METADATA_TYPE_PARAMETER = 0; // Parameter metadata
    METADATA_TYPE_EVENTS = 1; // Event definitions
    METADATA_TYPE_ACTUATORS = 2; // Actuator definitions
```

protos/component_metadata_server/component_metadata_server.proto
```diff
@@ -0,0 +1,32 @@
...
+
+message Metadata {
+    MetadataType type = 1; // The metadata type
+    string json_metadata = 2; // The JSON metadata
+}
+
+enum MetadataType {
+    PARAMETER = 0; // Parameter metadata
+    EVENTS = 1; // Event definitions
+    ACTUATORS = 2; // Actuator definitions
```
julianoes, line 31: Same here.

protos/component_metadata_server/component_metadata_server.proto
julianoes: So hang on, but you have to call it multiple times for the different metadata types, right?

julianoes (commented):
Couple of small things but pretty much there I think.

protos/component_metadata_server/component_metadata_server.proto
bkueng: No, the metadata is provided as an array: `repeated Metadata metadata`

protos/component_metadata/component_metadata.proto
bkueng: I added them

julianoes (approved):

PR #346, opened 2024-06-12
Events plugin

bkueng:
Proto file for https://github.com/mavlink/MAVSDK/pull/2338


julianoes (approved):
I think that API makes sense like this! I don't like the gazillion log levels (because I would never know how to classify things) but I assume that's given by the MAVLink implementation anyway, so that's fine.

PR #347, opened 2024-06-18
Updated proto `pb_plugin`

ShafSpecs:
Whilst attempting to fix issues within MAVSDK-Python outlined [here](https://github.com/mavlink/MAVSDK-Python/issues/698), I noticed dependency conflicts arising from the fact that the `pb_plugin` utilises very old versions of `protobuf`.

Updated those dependencies and formatted the project. Also fixed a few failing tests

### DIFF
--- a/pb_plugins/protoc_gen_mavsdk/autogen.py
+++ b/pb_plugins/protoc_gen_mavsdk/autogen.py
@@ -1,12 +1,12 @@
 # -*- coding: utf-8 -*-
 from os import environ
 from google.protobuf.compiler import plugin_pb2
-from .autogen_file import File
-from .docs import Docs
-from .enum import Enum
-from .methods import Method
-from .struct import Struct
-from .utils import (get_template_env,
+from protoc_gen_mavsdk.autogen_file import File
+from protoc_gen_mavsdk.docs import Docs
+from protoc_gen_mavsdk.pb_enum import Enum
+from protoc_gen_mavsdk.methods import Method
+from protoc_gen_mavsdk.pb_struct import Struct
+from protoc_gen_mavsdk.utils import (get_template_env,
                     has_result,
                     is_server,
                     name_parser_factory,
--- a/pb_plugins/protoc_gen_mavsdk/autogen.py
+++ b/pb_plugins/protoc_gen_mavsdk/autogen.py
@@ -7,10 +7,10 @@ from protoc_gen_mavsdk.pb_enum import Enum
 from protoc_gen_mavsdk.methods import Method
 from protoc_gen_mavsdk.pb_struct import Struct
 from protoc_gen_mavsdk.utils import (get_template_env,
-                    has_result,
-                    is_server,
-                    name_parser_factory,
-                    type_info_factory)
+                                     has_result,
+                                     is_server,
+                                     name_parser_factory,
+                                     type_info_factory)
 
 
 class AutoGen(object):
--- a/pb_plugins/protoc_gen_mavsdk/autogen.py
+++ b/pb_plugins/protoc_gen_mavsdk/autogen.py
@@ -12,7 +12,6 @@ from .utils import (get_template_env,
                     name_parser_factory,
                     type_info_factory)
 
-import protoc_gen_mavsdk.mavsdk_options_pb2
 
 class AutoGen(object):
     """ Autogenerator for the MAVSDK bindings """
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.egg-info/
 *__pycache__/
 *.pyc
+.idea/
 pb_plugins/generated/
 build/
 dist/
--- a/pb_plugins/protoc_gen_mavsdk/__init__.py
+++ b/pb_plugins/protoc_gen_mavsdk/__init__.py
@@ -2,4 +2,4 @@
 """ Autogenerator for MAVSDK """
 
 
-from .autogen import AutoGen
+from protoc_gen_mavsdk.autogen import AutoGen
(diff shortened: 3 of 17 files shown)


ShafSpecs:
Not sure if this is just my IDE going haywire or is a legit issue, but `google.protobuf.compiler.plugin_pb2` seem to lack any exports
https://github.com/mavlink/MAVSDK-Proto/blob/34bce6d61dba5f72b602bdbd31f04fceff093881/pb_plugins/protoc_gen_mavsdk/__main__.py#L12
Even though the [protobuf plugin header file](https://github.com/protocolbuffers/protobuf/blob/2088bc6e66850aa02cceaaae6096f39d20f1c677/src/google/protobuf/compiler/plugin.h#L53) seems to contain it

ShafSpecs:
> Not sure if this is just my IDE going haywire or is a legit issue, but `google.protobuf.compiler.plugin_pb2` seem to lack any exports
> 
> https://github.com/mavlink/MAVSDK-Proto/blob/34bce6d61dba5f72b602bdbd31f04fceff093881/pb_plugins/protoc_gen_mavsdk/__main__.py#L12
> 
> 
> Even though the [protobuf plugin header file](https://github.com/protocolbuffers/protobuf/blob/2088bc6e66850aa02cceaaae6096f39d20f1c677/src/google/protobuf/compiler/plugin.h#L53) seems to contain it

It looks like a breaking change from protobuf. `CodeGeneratorResponse` is no longer exported from compiler

JonasVautherin (approved):
I haven't tried it but it looks good to me :innocent:

ShafSpecs:
Added a new commit to fix circular imports bugs + removed relative import. Renamed a few files as well (`enum.py -> `pb_enum.py` as well as `struct.py`). Internal imports were resolving to the local files for some reason 👍

julianoes (approved):
I don't know much about this all but from far away it passes the squint test :).

I would say we merge this and the dependent PRs towards MAVSDK v3.

julianoes:
@ShafSpecs https://github.com/mavlink/MAVSDK/issues/2316

ShafSpecs:
Nice 🚀!

- Going through the code, my biggest pain point and something I would like to see in v3 is non-blocking asynchronous subscription. Not sure if this a server or MAVSDK-Python issue, but the hanging issue I raised [here](https://github.com/mavlink/MAVSDK-Python/issues/698#issuecomment-2186424142) (with status text monitor) where the generator hangs till a message is sent instead of having a background pub/sub architecture. With mavlink, it's first come, first serve, some messages get sent on a regular basis (heartbeat) whilst others randomly (statustext).
- And the server rpc stream not being synchronous in terms of how messages are queued, when a lot of plugins are subscribed to, errors about non-complete messages start to appear.
- I see someone has also raised the udp/tcp issue too for uniformity 👍 

My cpp isn't too sharp, so I am still studying the codebase (albeit slowly) but I have a hunch that the communication process can be improved to allow for more seamless data transfer. Thx

JonasVautherin:
> Not sure if this a server or MAVSDK-Python issue, but the hanging issue I raised https://github.com/mavlink/MAVSDK-Python/issues/698#issuecomment-2186424142 (with status text monitor) where the generator hangs till a message is sent

Hmm it's an async generator in asyncio, it is supposed to "hang" until it receives an event. Not sure if I understand your point correctly. What else would you have an async generator do? Until it returns, it has to "hang" until it receives the next event 🤔.

If you need to run multiple async generators in parallel, you can put them in their own coroutine, like e.g. [in the mission example](https://github.com/mavlink/MAVSDK-Python/blob/main/examples/mission.py#L19-L20).

I don't know why there was a `StatusCode.UNAVAILABLE` in [this particular issue](https://github.com/mavlink/MAVSDK-Python/issues/698#issuecomment-2186424142), though. In the code you share there, you seem to create an async generator for each event, which under the hood creates a new connection every time (instead of leveraging the gRPC streams, I would say). So maybe after some amount of connections it just stops answering (maybe some kind of backpressure, or maybe the connections are not getting closed correctly under the hood?). But in any case it feels suboptimal to use an async generator like this 🤔.

ShafSpecs:
> In the code you share there, you seem to create an async generator for each event, which under the hood creates a new connection every time (instead of leveraging the gRPC streams, I would say).

How would I leverage gRPC streams in that kind of situation (without resorting to new generators for every event? 

Looking at the v3 milestone, most of my suggestions are already penned down, that would be all from me ❤️

JonasVautherin:
> How would I leverage gRPC streams in that kind of situation (without resorting to new generators for every event?

I haven't seen your code base, but my guess is that you want to poll for the drone state (instead of receiving events in the async generator).

What I would do in that case is some kind of `Collector` that would subscribe to all the streams I want and that would save the latest state. At the same time it would expose an interface allowing to get the latest state.

Some quick pseudo-code:

```
class Collector {
    health = None
    position = None

    async def run_observers():
        asyncio.ensure_future(observe_health())
        asyncio.ensure_future(observe_position())

    async def observe_health():
        async for new_health in drone.telemetry.health():
            health = new_health

    async def observe_position():
        async for new_position in drone.telemetry.position():
            position = new_position
}
```

From the outside you run `Collector.run_observers()` in a coroutine, which will continuously update the state of `Collector`. And when you want to access the latest position, you read `Collector.position`.

Would that help?

julianoes:
Let's try it.

pb_plugins/protoc_gen_mavsdk/autogen.py
```diff
@@ -1,18 +1,17 @@
...
-from .enum import Enum
-from .methods import Method
-from .struct import Struct
-from .utils import (get_template_env,
-                    has_result,
-                    is_server,
-                    name_parser_factory,
-                    type_info_factory)
-
-import protoc_gen_mavsdk.mavsdk_options_pb2
```
julianoes, line 15: Why was this removed? :thinking:

PR #349, opened 2024-07-13
camera-server: Add ability to support taking image in video mode and video in image mode

jnomikos:
PR associated with this change:
https://github.com/mavlink/MAVSDK/pull/2351

### DIFF
--- a/protos/camera_server/camera_server.proto
+++ b/protos/camera_server/camera_server.proto
@@ -125,12 +125,6 @@ service CameraServerService {
 
     // Respond to an incoming tracking off command.
     rpc RespondTrackingOffCommand(RespondTrackingOffCommandRequest) returns(RespondTrackingOffCommandResponse) { option (mavsdk.options.async_type) = SYNC; }
-
-    // Set support for capturing images while in video mode. Affects CAMERA_CAP_FLAGS flags sent in CAMERA_INFORMATION.
-    rpc SupportImageInVideoMode(SupportImageInVideoModeRequest) returns(SupportImageInVideoModeResponse) { option (mavsdk.options.async_type) = SYNC; }
-
-    // Set support for capturing video while in image mode. Affects CAMERA_CAP_FLAGS flags sent in CAMERA_INFORMATION.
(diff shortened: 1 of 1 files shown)


JonasVautherin (commented):
Couldn't those go in the `SetInformation` function above? Since it affects the CAMERA_INFORMATION message :thinking:

jnomikos:
> Couldn't those go in the `SetInformation` function above? Since it affects the CAMERA_INFORMATION message 🤔

Yeah I could put it in `SetInformation`. The reason I didn't was because it is probably rare for a user to want to set this. But I agree it makes more sense to be there.

protos/camera_server/camera_server.proto
```diff
@@ -515,6 +521,20 @@ message RespondTrackingOffCommandResponse {
...
 
+message SupportImageInVideoModeRequest {
+    bool support = 1; // True if the camera supports capturing images while in video mode
+}
+message SupportImageInVideoModeResponse {
+    CameraServerResult camera_server_result = 1;
+}
+
+message SupportVideoInImageModeRequest {
+    bool support = 1; // True if the camera supports capturing video while in image mode
```
julianoes, line 532:
```suggestion
    bool is_supported = 1; // True if the camera supports capturing video while in image mode
```

protos/camera_server/camera_server.proto
```diff
@@ -515,6 +521,20 @@ message RespondTrackingOffCommandResponse {
     CameraServerResult camera_server_result = 1; // The result of sending the response.
 }
 
+message SupportImageInVideoModeRequest {
+    bool support = 1; // True if the camera supports capturing images while in video mode
```
julianoes, line 525:
```suggestion
    bool is_supported = 1; // True if the camera supports capturing images while in video mode
```

julianoes (commented):
I think that makes sense. It would be nice if this came implicitly out of the camera subscriptions but I don't think there is an easy way to do so.

And thanks for the PR!

julianoes (approved):
I think that's easier and makes more sense.

PR #351, opened 2024-07-31
Gimbal API redesign (support more than one gimbal, support streams and commands)

julianoes:
- Adding an argument for send mode (either command or message stream)
- Includes the mode into the setter because that's when it's sent.  Otherwise, the set_mode call itself doesn't have any effect which is not very intuitive.
- Extend interface to support multiple gimbals.


JonasVautherin (approved):
Looks good to me!

julianoes:
@JonasVautherin thanks for the review!

PR #352, opened 2024-08-29
camera: change API to support multiple cameras

julianoes:
This changes the API to support more than one camera within one camera plugin instance.

This will enable multiple cameras in language wrappers instead of just C++ as it is now.

### DIFF
--- a/protos/camera/camera.proto
+++ b/protos/camera/camera.proto
@@ -18,10 +18,6 @@ option java_outer_classname = "CameraProto";
  */
 service CameraService {
     /*
-     * Prepare the camera plugin (e.g. download the camera definition, etc).
-     */
-    rpc Prepare(PrepareRequest) returns(PrepareResponse) {}
-    /*
      * Take one photo.
      */
     rpc TakePhoto(TakePhotoRequest) returns(TakePhotoResponse) {}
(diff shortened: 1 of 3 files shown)


protos/camera/camera.proto
```diff
@@ -17,10 +17,6 @@ option java_outer_classname = "CameraProto";
  * `select_camera`.
  */
 service CameraService {
-    /*
-     * Prepare the camera plugin (e.g. download the camera definition, etc).
-     */
-    rpc Prepare(PrepareRequest) returns(PrepareResponse) {}
```
JonasVautherin, line 23: Why is this being removed? The idea was to be able to download the camera definition in advance, because it may be slow/timeout.

protos/camera/camera.proto
```diff
@@ -57,34 +53,57 @@ service CameraService {
...
+     *
+     * This allows to find out what cameras are connected to the system.
+     * Based on the camera ID, we can then address a specific camera.
+     */
+    rpc SubscribeCameraList(SubscribeCameraListRequest) returns(stream CameraListResponse) {}
     /*
      * Subscribe to camera mode updates.
      */
-    rpc SubscribeMode(SubscribeModeRequest) returns(stream ModeResponse) {}
+    rpc SubscribeMode(SubscribeModeRequest) returns(stream ModeResponse) { option (mavsdk.options.async_type) = ASYNC; }
```
JonasVautherin, line 66: I am not sure I understand the change putting all the streams as ASYNC and adding a Get* version for all of them. Isn't it already the case that the SYNC version is blocking (but it is called `mode()` instead of `get_mode()`)? E.g. https://github.com/mavlink/MAVSDK/blob/main/src/mavsdk/plugins/camera/include/plugins/camera/camera.h#L675

protos/camera/camera.proto
julianoes: My plan is to have it initialize automatically when it discovers a camera.

protos/camera/camera.proto
julianoes: So there is a catch with multiple cameras.

If you leave it the way it is, you get subscriptions for all cameras, as they get updated. However, when you call `mode()` which one do you get? To avoid that, I remove `mode()` and add `get_mode(camera_id)`.

protos/camera/camera.proto
JonasVautherin: Hmm... IIRC we've had issues in the past where is was being slow and either blocking stuff or producing a lot of timeouts. Then it was lazily initialized, but it caused issues because the very first call to a camera function was sometimes super slow (again, timeouts...).

Making it explicit has a few advantages:
* If you don't use the camera at all, you don't download the camera definition
* You can download the camera definition before you start using the camera (because usually you know in advance that you will use it)

protos/camera/camera.proto
JonasVautherin: Oh, of course!

protos/camera/camera.proto
julianoes: With C++ you would actually not even instantiate the plugin but I can see that for mavsdk_server it's a different story.

Beat added file caching for MAVSDK, so if we use that to cache the FTP file, it should be fairly quick actually.

JonasVautherin: Your call :+1:. I just wanted to mention why I added this function (in my experience the camera definition has always been the root of tons of camera problems)

julianoes: Right, I hear you. Let me try and see how it goes.

PR #355, opened 2024-10-28
action: remove max speed settings

julianoes:
This API is confusing as it only works for PX4 quads but not PX4 fixedwing or anything ArduPilot based.

The reason is that this just sets a PX4 parameter that is not standardized via MAVLink.

Therefore, I suggest to remove the API and instead recommend to use the param plugin to set the specific params specifically.


JonasVautherin (approved):
Totally agreed!

PR #357, opened 2024-12-09
API for receiving status notifications on outgoing mission requests

jonathanreeves:
This is a companion PR for the main implementation in MAVSDK:
https://github.com/mavlink/MAVSDK/pull/2463

Details can be found in the MAVSDK PR.

### DIFF
--- a/protos/mission_raw_server/mission_raw_server.proto
+++ b/protos/mission_raw_server/mission_raw_server.proto
@@ -16,6 +16,11 @@ service MissionRawServerService {
     rpc SubscribeIncomingMission(SubscribeIncomingMissionRequest) returns(stream IncomingMissionResponse) {}
 
     /*
+     * Subscribe to when a new mission download request completes (asynchronous)
+     */
+    rpc SubscribeOutgoingMission(SubscribeOutgoingMissionRequest) returns(stream OutgoingMissionResponse) { option (mavsdk.options.async_type) = ASYNC; }
+
+    /*
      * Subscribe to when a new current item is set
      */
     rpc SubscribeCurrentItemChanged(SubscribeCurrentItemChangedRequest) returns(stream CurrentItemChangedResponse) {}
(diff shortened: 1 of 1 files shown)


julianoes (approved):

julianoes:
This is ok, but my question is: do you need this notification? All you care about is if the mission is uploaded, so you are given a new mission, right? You don't really need to know when someone downloads it, I would think, it's more of an internal thing.

JonasVautherin:
> do you need this notification?

Isn't this the API to download a mission? I.e. the proto for https://github.com/mavlink/MAVSDK/pull/2463?

jonathanreeves:
> This is ok, but my question is: do you need this notification? All you care about is if the mission is uploaded, so you are given a new mission, right? You don't really need to know when someone downloads it, I would think, it's more of an internal thing.

Yeah it's a great question, and one that I've definitely gone back and forth on. I think you could make a weak argument that app code would want the _option_ to know if something went wrong during an air -> ground mission download, but if it succeeds you probably don't care. Even still it could be viewed as an internal thing that's happening behind the scenes, in some ways it's cleaner for users of the SDK to not have to worry about it. Honestly I'm very comfortable throwing these changes away if you think that's best for the API. I was sort of just erring on the side of being explicit and make sure error information wasn't hidden. Just let me know your thoughts, I will not at all be offended if we think the best thing is just close this and keep the error status hidden behind the scenes.

jonathanreeves:
> > do you need this notification?
> 
> Isn't this the API to download a mission? I.e. the proto for [mavlink/MAVSDK#2463](https://github.com/mavlink/MAVSDK/pull/2463)?

It is, but my guess is this is error information that most users of the SDK (myself including honestly) are not super likely to use. There's no application interaction required for a ground station to request and stream a mission plan from the air vehicle.

julianoes:
I would say that if you are using it or intending on using it, then that means there is a use case and you should leave it. If not, I'd probably pass on it.

jonathanreeves:
> I would say that if you are using it or intending on using it, then that means there is a use case and you should leave it. If not, I'd probably pass on it.

I doubt we'll be using it anytime soon so I'll go ahead and close this. I'm still planning to open a proto PR with the changes to make these methods async (as discussed in the mission download change), but we can deal with that separately.

PR #361, opened 2025-01-19
mocap: add SetVisionSpeedEstimate

Luc-Meunier:
Hi, we would like to add the VISION_SPEED_ESTIMATE (103) MAVLink message to the mocap proto.

See https://mavlink.io/en/messages/common.html#VISION_SPEED_ESTIMATE

Relates to issue mavlink/MAVSDK#2497


protos/mocap/mocap.proto
```diff
@@ -105,6 +114,13 @@ message VisionPositionEstimate {
     Covariance pose_covariance = 4; // Pose cross-covariance matrix.
 }
 
+// Global speed estimate from a vision source.
+message VisionSpeedEstimate {
+    uint64 time_usec = 1; // PositionBody frame timestamp UNIX Epoch time (0 to use Backend timestamp)
+    SpeedBody speed_body = 2; // Global speed (m/s)
```
julianoes, line 120: Just wondering on the frame here? When you see "Global" do you mean in North, East, Down, or Forward, Left, Down? I'm pretty sure "Body" would imply in body frame, so Forward, Left, Down.

julianoes (commented):
Thanks for the pull request! Just one question regarding a comment.

protos/mocap/mocap.proto
Luc-Meunier: Hi @julianoes! Thansk for the review!

I used the same description as in the MAVLink doc, but I guess this is a typo.

However, I see in the VisionPositionEstimate a similar case:
`PositionBody position_body = 2; // Global position (m)`
Should I fix it too?

julianoes: Interesting. Might be worth checking what PX4 expects and whether we should update the MAVLink messages description.

julianoes: For position, PX4 expects NED:

https://github.com/PX4/PX4-Autopilot/blob/1eb9434b8c39fef3691684df520686231dce3939/src/modules/mavlink/mavlink_receiver.cpp#L1319C40-L1322

I don't see vision speed being used in PX4. How are you testing this?

julianoes: Interesting, so ArduPilot implements it as body. How does it implement position updates? Also body? In which case we have found yet another inconsistency between PX4 and ArduPilot unfortunately... And need to deal with it in MAVLink and/or MAVSDK somehow.

fed12345: Hi all,
to clear this up a bit, if we follow the mocap position from MAVSDK.
1)  The mocap function set_vision_position_estimate() expects global position(NED) according to [documentation](https://mavsdk.mavlink.io/v2.0/en/cpp/api_reference/classmavsdk_1_1_mocap.html#classmavsdk_1_1_mocap_1a22d007409839e28a45d7b10f10e22fd6) the input parameter for this function calls the position [PositionBody](https://mavsdk.mavlink.io/v2.0/en/cpp/api_reference/structmavsdk_1_1_mocap_1_1_vision_position_estimate.html) so this might be a typo.
2) MAVSDK uses the mavlink [VISION_POSITION_ESTIMATE](https://github.com/mavlink/MAVSDK/blob/main/src/mavsdk/plugins/mocap/mocap_impl.cpp#L62) message which according to [documentation](https://mavlink.io/en/messages/common.html#VISION_POSITION_ESTIMATE) expects the local frame(FRD, not body).
3) Ardupilot does not differentiate between local(FRD) and global(NED) frame as it doesn't have a local frame, so it interprets the position in global frame. In fact, it processes the messages VISION_POSITION_ESTIMATE and GLOBAL_VISION_POSITION_ESTIMATE the same way, see [here](https://github.com/ArduPilot/ardupilot/blob/master/libraries/GCS_MAVLink/GCS_Common.cpp#L3837)

Ardupilot also expects global(NED) position from the mocap class, and consequently global(NED) velocity. So, if PX4 is the same, I think the only thing that needs to be changed is the variable's name in the [VisionPositionEstimate](https://mavsdk.mavlink.io/v2.0/en/cpp/api_reference/structmavsdk_1_1_mocap_1_1_vision_position_estimate.html) structure, as @Luc-Meunier suggested. 
Let me know if you agree

julianoes: Ok, so if both PX4 and ArduPilot interpret the values as NED, that's good news. In that case I'm tempted to rename the `PositionBody` type to `PositionNed` to make it clear. And then we should update the MAVSDK and MAVLink docs to make it clear.

Luc-Meunier: Thanks @fed12345 and @julianoes, I added a SpeedNed struct.

miodine: Hello,

Just as a quick follow-up to this:
> I'm tempted to rename the PositionBody type to PositionNed to make it clear...

To us it also seems like a good idea, should we do it in another PR or would it fly to add it in this one?  
Aside from that, is there anything else that we should take care of?

julianoes: I looked into it but was very confused by the fact that ArduPilot just handles local and global the same way. That doesn't really make sense to me, but anyway, I think you're right that both handle it NED, and so we should probably rename the field `PositionBody` to `PositionNed`.

> Ardupilot does not differentiate between local(FRD) and global(NED) frame as it doesn't have a local frame, so it interprets the position in global frame. In fact, it processes the messages VISION_POSITION_ESTIMATE and GLOBAL_VISION_POSITION_ESTIMATE the same way, see [here](https://github.com/ArduPilot/ardupilot/blob/master/libraries/GCS_MAVLink/GCS_Common.cpp#L3837)

It would have to differentiate between FRD and NED, as they are not the same! But you are saying it just does NED anyway? I couldn't tell from the code snippet.

fed12345: The code snippet only shows that whether ardupilot receives the `VISION_POSITION_ESTIMATE` or the `GLOBAL_VISION_POSITION_ESTIMATE` message from Mavlink, it still processes the message with the function `handle_common_vision_position_estimate_data`, so it doesn't care if it's global or local. 

I also agree that this whole thing is confusing. This is how I interpret the frames:

Global: Latitude and longitude, GPS stuff
NED: Origin at initialization of Kalman Filter, yaw aligned with North
Local FRD: Origin at initialization of Kalman Filter, yaw aligned with front of the drone
Body FRD: Origin at cog of drone, yaw aligned with front of the drone

Ardupilot does not have a concept of Local FRD only body FDR.

fed12345:
Hi, do we have any news on this?

protos/mocap/mocap.proto
julianoes: Sorry, bit late here.

Yes, I'm usually thinking of body FRD as well.

So, I'm still confused what ArduPilot expects here:

https://github.com/ArduPilot/ardupilot/blob/7a1630c93f5887c40ef43fa247cfa856dfdd19ba/libraries/GCS_MAVLink/GCS_Common.cpp#L3988

Is it Local NED, Body NED, or Body FRD? From that we should be able to name and comment the fields here.

As said above, I believe PX4 wants Local NED.

fed12345: Ardupilot also wants Local NED

julianoes: Ok good. And you already renamed it accordingly. Sorry, I missed that. That's all good then.

julianoes (approved):

julianoes:
We'll merge this PR once the PR for MAVSDK is ready.

Luc-Meunier:
> We'll merge this PR once the PR for MAVSDK is ready.

Hi, here is a PR for MAVSDK: https://github.com/mavlink/MAVSDK/pull/2538

julianoes (approved):

PR #363, opened 2025-01-23
Bump to 1.2.0

JonasVautherin:
This is needed for deployment, otherwise it tries to overwrite 1.1.1 and fails (see [here](https://github.com/mavlink/MAVSDK-Proto/actions/runs/12862022784/job/36055095760)) 👍.


julianoes (approved):
Ah, thanks!

julianoes:
Ah, but it will be 1.2.1 nex time, right?

PR #366, opened 2025-02-12
We need mavsdk_options_pb2 to stay

JonasVautherin:
From `pb_plugins/`, I re-generated `mavsdk_options_pb2.py` with a newer `protoc` with:

```
protoc -I../protos --python_out=./protoc_gen_mavsdk ../protos/mavsdk_options.proto
```

And imported it into `autogen.py`, similar to how it was before #347.

@julianoes does that help? I have only tried to generate a few things and the generation step was fine. I haven't tried to build anything further than that, but if that fails feel free to give me instructions to reproduce :blush: :+1:.


julianoes:
Looks like this fixes it, thanks!

PR #370, opened 2025-02-18
ActionServer: allow direct setting of arm/disarm state and vehicle flight mode

jonathanreeves:
# Summary
The `ActionServer` implementation currently only supports updating arm/disarm status and flight mode by receiving MAVLINK messages from a client. An autopilot implementation may need to take autonomous action to change flight modes or report an armed status apart from explicit GCS command. This PR provides the necessary support to do that.


JonasVautherin (approved):
Nice!

protos/action_server/action_server.proto
```diff
@@ -55,6 +59,14 @@ message SetAllowableFlightModesRequest
     AllowableFlightModes flight_modes = 1;
 }
 
+message SetArmedStateRequest {
+    bool is_armed = 1; // Is armed now?
+}
+
+message SetFlightModeRequest {
+    FlightMode flight_mode = 1;
```
julianoes, line 67: Probably needs a docstring.

julianoes (commented):
Thanks!

protos/action_server/action_server.proto
jonathanreeves: Agreed, makes sense. I'll add it right now.

jonathanreeves: Done

julianoes (approved):

PR #375, opened 2025-05-20
Info: Disambiguate method signature

matthieu-db:
Having the same method name made compiling the java generated code fail


JonasVautherin:
The problem is that this will break the MAVSDK API everywhere. I think a better solution would be for the Java autogeneration templates to handle this case properly (seems like the other languages are fine, right?).

What do you think?

JonasVautherin:
Yeah in my experience it's only Java. I think until now I've just removed the function in the proto file before generating the Java code (nobody ever complained about this feature missing there 😅)

PR #405, opened 2026-04-06
Move HomePosition after Quaternion in telemetry proto

PavelGuzenfeld:
## Summary
- Reorder `HomePosition` message to appear after `Quaternion` in telemetry.proto

## Motivation
The C++ code generator (`protoc-gen-mavsdk`) preserves proto message order when generating struct definitions. `HomePosition` contains a `Quaternion q` field, so it must appear after `Quaternion` in the proto file — otherwise the generated C++ header has a forward-dependency error (`'Quaternion' does not name a type`).

This was introduced in #399 where `HomePosition` was added between `Position` and `Heading`, but `Quaternion` is defined later.

## Changes
- Move `HomePosition` message block to after `Quaternion` (no field or numbering changes)

Companion fix: https://github.com/mavlink/MAVSDK/pull/2839

### DIFF
--- a/protos/telemetry/telemetry.proto
+++ b/protos/telemetry/telemetry.proto
@@ -499,27 +499,6 @@ message Position {
     float relative_altitude_m = 4 [(mavsdk.options.default_value)="NaN"]; // Altitude relative to takeoff altitude in metres
 }
 
-/*
- * Home position type.
- *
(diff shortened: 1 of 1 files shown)


julianoes (approved):
Aha!

PR #414, opened 2026-06-10
Ability to set camera_server positional info, zoom factor and fov

jnomikos:
Adds support for setting camera_server position, attitude, zoom factor and field of view.

This is to add CAMERA_FOV_STATUS support in this PR here: https://github.com/mavlink/MAVSDK/pull/2898


julianoes:
Taken your commits and finished it here: #419.

PR #420, opened 2026-07-24
Add marine plugins

gisleandre:
## Marine plugins: autonomy, payload, and subsea telemetry

Adds proto definitions for maritime/subsea vehicles (USVs and UUVs). Four new
plugins plus additions to the existing telemetry plugins.

#### Autonomy / Autonomy Server
Control and monitoring of an onboard autonomy system. The client side queries
`AutonomyStatus` (state machine: idle / executing / paused / suspended /
completed / stopped / recovering) and issues `AutonomyDemand` transitions
(start, stop, pause, suspend, resume, go-to-recovery). The server side sets
state, streams incoming demands so the autonomy system can accept, deny, or
temporarily reject them, and reports per-component health for the subsystems
typically present on a subsea vehicle (ATR, SAS, MBES, FLS, camera, acoustic
comms, mission).

#### Payload / Payload Server
Registration and control of vehicle payloads. Payloads are enumerated with an
ID, name, and type (camera, LiDAR, winch, acoustic modem, SAS, MBES, FLS, plus
five generic slots), and each exposes a powered / armed / active state along
with a health flag. Clients list payloads, read status, and request state
changes; servers add and remove payloads and publish their status.

#### WaterCurrent and DopplerVelocityLog (DVL)
New messages with subscribe, set-rate, and publish calls in both plugins. Water
current covers raw and estimated current in NED plus sensor distance; DVL
covers bottom-lock, per-beam seabed altitude, and vehicle velocity in NED.

#### Wind
Adds `timestamp_us` (new field number, wire-compatible), the missing
`SetRateWind`, and the message plus `PublishWind` on the server side, which
previously had no way to publish wind.

### DIFF
--- a/protos/payload/payload.proto
+++ b/protos/payload/payload.proto
@@ -1,0 +1,80 @@
+syntax = "proto3";
+
+package mavsdk.rpc.payload;
+
+import "mavsdk_options.proto";
+
+option java_package = "io.mavsdk.payload";
+option java_outer_classname = "PayloadProto";
+
+// Query payload information.
+service PayloadService {
+
+    /*
+     * Get the list of all connected payloads.
+     */
+    rpc GetPayloadList(GetPayloadListRequest) returns(GetPayloadListResponse) {}
+
+    /*
+     * Subscribe to payload updates.
+     */
+    rpc SubscribePayloadUpdate(SubscribePayloadUpdateRequest) returns(stream PayloadUpdateResponse) { option (mavsdk.options.async_type) = SYNC; }
+}
+
+enum PayloadType {
+    PAYLOAD_CAMERA = 0; // Generic Camera Payload
+    PAYLOAD_LIDAR = 1; // Light Detection and Ranging Payload
+    PAYLOAD_WINCH = 2; // Winch Payload
+    PAYLOAD_ACOUSTIC_MODEM  = 3; // Acoustic Modem Payload
+    PAYLOAD_SAS  = 4; // Single Aperture Sonar Payload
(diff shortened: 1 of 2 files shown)


julianoes:
Thanks. Looks like this would be pending the outcome of https://github.com/mavlink/mavlink/pull/2556.

hamishwillee:
> Thanks. Looks like this would be pending the outcome of [mavlink/mavlink#2556](https://github.com/mavlink/mavlink/pull/2556).

FWIW that PR adds the dialect _as a dialect_ to the mavlink repo - not as a managed definitions such as those in common.xml. For my information, what's MAVSDK approach for supporting non-standard definitions?

julianoes:
MAVSDK will try to support what is out there, e.g. have a few quirks for PX4 and ArduPilot in order to be compatible, but MAVSDK likely won't add support for non-standard dialects for new/future messages.

hamishwillee:
Makes sense. I assume though someone can add their own MAVSDK external plugins in their own repo should they wish?

julianoes:
> I assume though someone can add their own MAVSDK external plugins in their own repo should they wish?

It involves a fork as it's really modules and not plugins (they don't plug in). So yes, possible but annoying.

What I would say is that this functionality can easily be implemented on the library user side instead using MavlinkDirect.

julianoes:
https://github.com/mavlink/mavlink/pull/2556 was closed, closing this as well.

If @gisleandre comes back and responds to our questions, I'm happy to open this up again.

PR #421, opened 2026-07-28
core: raise heartbeat watchdog minimum timeout to 2 s

julianoes:
Follow-up to #415, which added the heartbeat watchdog RPCs.

MAVSDK sends its periodic heartbeats at 1 Hz and only evaluates the watchdog deadline when a heartbeat is due. With a 1 s timeout, a feed and the tick can end up almost in phase, so a single slightly late feed already drops a heartbeat and leaves a 2 s gap on the wire.

This showed up concretely while reviewing the implementation side: a test using a 1 s timeout with a single feed failed 2 of 3 runs purely on feed/tick phase, and every test had to be moved off 1 s to be stable. A documented minimum that cannot actually be used is a trap, so raise it to two heartbeat periods.

Comment-only change, so no generated code changes. `prototool lint protos` passes with the same image CI uses.

Changes:

- `SetHeartbeatWatchdogTimeout`: values below 2 are rejected, with the reason.
- `SetHeartbeatWatchdogTimeoutRequest.timeout_s`: minimum 2 rather than 1.
- `FeedHeartbeatWatchdog`: say to feed at least *twice* per timeout period. "At least once per timeout period" reads as sufficient but leaves no margin for jitter.

The implementation change is in mavlink/MAVSDK#2895, which pins this repo as a submodule — that PR needs this one merged first so it can bump the submodule to a commit whose documented minimum matches the enforced one.


julianoes:
@tpayne-censystech does this make sense? When we have the watchdog at 1 Hz, it can get out of sync and "just miss it".

tpayne-censystech:
@julianoes yeah, good catch. It's effectively an aliasing issue. The heartbeat watchdog timeout should be at least double the heartbeat period to prevent "aliasing".

PR #427, opened 2026-08-24
shell: add SetDevice for SERIAL_CONTROL_DEV selection

Bartok9:
## Summary
Add `SetDevice` so the shell plugin can target MAVLink `SERIAL_CONTROL_DEV` ports beyond the default shell device (e.g. TELEM2).

## Linked
- mavlink/MAVSDK#2107

## Notes
Companion MAVSDK implementation follows this proto change.


protos/shell/shell.proto
```diff
@@ -21,6 +26,14 @@ service ShellService {
...
+     * Select the SERIAL_CONTROL device used for send/receive.
+     *
+     * Default is DEVICE_SHELL. Other values map to MAVLink SERIAL_CONTROL_DEV
+     * (TELEM1/2, GPS1/2, SERIALn). Incoming SERIAL_CONTROL messages for other
+     * devices are ignored while a device is selected.
+     */
+    rpc SetDevice(SetDeviceRequest) returns(SetDeviceResponse) { option (mavsdk.options.async_type) = SYNC; }
 }
 
 message SendRequest {
```
julianoes: What about adding it to the request and response instead?

We can add/change arguments before releasing v4.

Bartok9: Good call — dropped `SetDevice` and put `Device` on `SendRequest` (+ `ReceiveResponse` so callers can see the source device). Default remains `DEVICE_SHELL`. Pushed in e743112.

Bartok9:
@julianoes thanks — updated per your note: removed `SetDevice` and added `Device` on `SendRequest` and `ReceiveResponse` (pre-v4 field add). Default remains `DEVICE_SHELL`.

julianoes (dismissed):

Bartok9:
@julianoes gentle ping — CI is green again on the latest tip.

Since the nested `Receive` + doxygen brief commits landed after your earlier approval, GitHub dismissed that review. Ready for another look when you have a moment (proto side only; MAVSDK#3042 stays draft on the matching tip).

protos/shell/shell.proto
```diff
@@ -23,16 +28,42 @@ service ShellService {
...
+    DEVICE_SERIAL0 = 100; // SERIAL_CONTROL_SERIAL0
+    DEVICE_SERIAL1 = 101; // SERIAL_CONTROL_SERIAL1
+    DEVICE_SERIAL2 = 102; // SERIAL_CONTROL_SERIAL2
+    DEVICE_SERIAL3 = 103; // SERIAL_CONTROL_SERIAL3
+    DEVICE_SERIAL4 = 104; // SERIAL_CONTROL_SERIAL4
+    DEVICE_SERIAL5 = 105; // SERIAL_CONTROL_SERIAL5
+    DEVICE_SERIAL6 = 106; // SERIAL_CONTROL_SERIAL6
+    DEVICE_SERIAL7 = 107; // SERIAL_CONTROL_SERIAL7
+    DEVICE_SERIAL8 = 108; // SERIAL_CONTROL_SERIAL8
+    DEVICE_SERIAL9 = 109; // SERIAL_CONTROL_SERIAL9
```
julianoes, line 47: The values look like they would make sense but it turns out the values are actually ignored in the language wrappers, from what I can tell from the generated code. @JonasVautherin do you remember how we should do this?

Bartok9: Thanks @julianoes — good catch on the wrapper side.

You're right that plain protobuf enum **numeric values** are often not what the language wrappers expose as a first-class API surface; `protoc-gen-mavsdk` typically maps the enum **names** into the generated C++/Python/etc. types, and the wire numbers only matter for the protobuf encoding.

I intentionally aligned the enum numbers with MAVLink `SERIAL_CONTROL_DEV` so the C++ plugin can do a straight cast/`static_cast` when filling `SERIAL_CONTROL.device` (and so the values stay stable if anyone inspects the raw proto). If that mapping is wrong for the generator / wrapper story, I'm happy to:

1. keep the **names** as-is and renumber to the usual 0-based MAVSDK style, with an explicit map in the C++ plugin (MAVSDK#3042), or  
2. follow whatever pattern @JonasVautherin recommends for enums that mirror MAVLink constants.

Happy to adjust as soon as you two land on the preferred approach — proto PR stays ready either way; companion MAVSDK#3042 stays draft until this lands.

julianoes: I don't think you use a static cast, do you? I think it's a switch case anyway?

Bartok9: You're right — I misspoke earlier. Companion MAVSDK#3042 maps with explicit `switch` cases (C↔C++ wrappers and shell_impl → SERIAL_CONTROL.device), not a static_cast on the proto enum numbers.

So matching MAVLink wire values in the proto enum is convenient but not required. Happy to renumber to the usual 0-based MAVSDK style in this PR if you prefer; the switch tables absorb either layout. Want me to push that?

julianoes: I'm just trying to decide what's more intuitive and less likely to go wrong for a user of C++ and/or Python...

julianoes: Decided this is ok, and we're changing the tooling in v4 to handle it correctly.

julianoes (approved):

julianoes (approved):

PR #429, opened 2026-09-14
protoc-gen-mavsdk: expose enum value numbers, release 2.1.0

julianoes:
## Summary

Enum values now carry the number from the proto as `value.number`, so templates can emit the actual value instead of `loop.index0`. Today every binding generated from these templates uses the declaration order, which silently diverges from the proto for enums that mirror MAVLink numbering (`MAV_FRAME` 0/8/16/18, `FAILURE_UNIT` 0..8/100..105, `StorageType` ..7/254, several `Result` enums with gaps). That bit us in MAVSDK: mavlink/MAVSDK#3076.

Two small fixes that let the MAVSDK C++ templates move off the pinned 1.x generator onto 2.x:

- `NameParser.__str__` returns the original name, so `{{ type_info.name }}` still renders message type names as a string, as it did in 1.x.
- `TypeInfo.is_result` no longer raises on repeated message types, where `name` is a string rather than a `NameParser`.

Version bumped to 2.1.0. The consuming change, with all bindings regenerated, is in mavlink/MAVSDK (PR to follow); its CI generation check needs 2.1.0 tagged and on PyPI to pass.

The existing `pb_plugins/test` failures (21) are pre-existing: those tests were never updated for the 2.0 `TypeInfoFactory` constructor. No new failures.

Context: mavlink/MAVSDK-Proto#427 review discussion about explicit enum values.


julianoes:
Gonna try that one, hopefully it'll work out.

JonasVautherin (approved):

